PR #1, opened 2022-01-30
New LLVM Backend

leissa:
This removes the old LLVM backend with a new LLVM backend that simply spits out a string without linking to LLVM.


leissa:
Although the new backend is far from being complete, I'm going to merge anyway. This allows to use CI much more easily.

PR #3, opened 2022-02-07
wip: thorin fe

leissa:


leissa:
This was more a check to see whether the `exec` thing works on msvc/windows.

PR #15, opened 2022-03-11
Add some initial Natvis file to make debugging thorin in VS Code less painful

fodinabor:
Shows the node type in string form directly for each Def. For Axioms, it shows the axiom's tag.
Additionally, the ops are shown as array elements. These are otherwise not inspectable "sanely" at all.
Can be greatly expanded, but this was what I needed today :)

Exemplary view:
image


leissa:
Nice. Can you add a new line at the end of `thorin.natvis`?

PR #16, opened 2022-03-14
Add affine for-loop axiom

fodinabor:
This adds a `For` axiom including a `LowerFor` rewrite pass that lowers the axiom to a CPS loop.
The axiom is defined following this scheme:
` for :: [m: Nat , n: Nat , Ts: «n; *»] → [Mem , Int m, Int m, Int m, «i: n; Is#i», Cn [Mem , «i: n; Is#i», Cn [Mem , «i: n; Is#i»]], Cn [Mem , «i: n; Is#i»]];`

Note, that due to 1. the design of the return cn (doesn't have a continuation itself) and 2. some _unexpected_ indirection, the LL backend currently requires all passes including `CopyProp` to run in an optimization phase after the lowering. Otherwise, assertion failures/crashes occur.

Happy about any feedback!

### DIFF
--- a/src/thorin/pass/rw/lower_for.cpp
+++ b/src/thorin/pass/rw/lower_for.cpp
@@ -1,7 +1,6 @@
 #include "thorin/pass/rw/lower_for.h"
 
 #include "thorin/lam.h"
-#include "thorin/tables.h"
 
 namespace thorin {
 
@@ -10,7 +9,7 @@ const Def* LowerFor::rewrite(const Def* def) {
         if (auto for_ax = app->axiom(); for_ax && for_ax->tag() == Tag::For) {
             auto& w = def->world();
             w.DLOG("rewriting for axiom: {} within {}", app, curr_nom());
-            
+
             auto for_pi  = app->callee_type();
             auto for_lam = w.nom_lam(for_pi, w.dbg("for"));
 
--- a/src/thorin/pass/rw/lower_for.cpp
+++ b/src/thorin/pass/rw/lower_for.cpp
@@ -39,7 +39,7 @@ const Def* LowerFor::rewrite(const Def* def) {
             auto if_else    = w.nom_lam(if_else_cn, nullptr);
             if_else->app(false, brk, {if_else->mem_var(w.dbg("mem")), acc});
 
-            auto cmp = w.op(ICmp::sl, iter, stop);
+            auto cmp = w.op(ICmp::ul, iter, stop);
             for_lam->branch(false, cmp, if_then, if_else, mem);
         }
 
--- a/src/thorin/pass/rw/lower_for.cpp
+++ b/src/thorin/pass/rw/lower_for.cpp
@@ -5,48 +5,46 @@
 namespace thorin {
 
 const Def* LowerFor::rewrite(const Def* def) {
-    if (auto app = def->isa<App>())
-        if (auto for_ax = app->axiom(); for_ax && for_ax->tag() == Tag::For) {
-            auto& w = def->world();
-            w.DLOG("rewriting for axiom: {} within {}", app, curr_nom());
+    if (auto for_ax = isa<Tag::For>(def)) {
+        auto& w = world();
+        w.DLOG("rewriting for axiom: {} within {}", for_ax, curr_nom());
 
-            auto for_pi  = app->callee_type();
-            auto for_lam = w.nom_lam(for_pi, w.dbg("for"));
+        auto for_pi  = for_ax->callee_type();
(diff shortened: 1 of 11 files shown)


leissa:
Nice work. I like it a lot :)

src/thorin/pass/rw/lower_for.cpp
```diff
@@ -0,0 +1,54 @@
+#include "thorin/pass/rw/lower_for.h"
+
+#include "thorin/lam.h"
+
+namespace thorin {
+
+const Def* LowerFor::rewrite(const Def* def) {
+    if (auto app = def->isa<App>())
+        if (auto for_ax = app->axiom(); for_ax && for_ax->tag() == Tag::For) {
+            auto& w = def->world();
```
leissa, line 10: There is already a `world()` getter defined in `Pass`.

src/thorin/pass/rw/lower_for.cpp
```diff
@@ -0,0 +1,54 @@
+#include "thorin/pass/rw/lower_for.h"
+
+#include "thorin/lam.h"
+
+namespace thorin {
+
+const Def* LowerFor::rewrite(const Def* def) {
+    if (auto app = def->isa<App>())
+        if (auto for_ax = app->axiom(); for_ax && for_ax->tag() == Tag::For) {
```
leissa, line 9: You can match that a bit simpler. Sth like this:
```cpp
if (auto for_ax = isa<Tag::For>(def)) /*...*/
```
This will give you a `Query` that lets you access the axiom via `Query::axiom` and otherwise behaves like the matched `def`.

src/thorin/pass/rw/lower_for.cpp
```diff
@@ -0,0 +1,54 @@
...
+            auto& w = def->world();
+            w.DLOG("rewriting for axiom: {} within {}", app, curr_nom());
+
+            auto for_pi  = app->callee_type();
+            auto for_lam = w.nom_lam(for_pi, w.dbg("for"));
+
+            auto body      = app->args()[app->num_args() - 2];
+            auto body_type = body->type()->as<Pi>();
+            auto cont_pi   = body_type->dom(body_type->num_doms() - 1)->as<Pi>();
```
leissa, line 18: maybe:
```cpp
auto cont_pi = body_type->doms().back()->as<Pi>();
```

src/thorin/pass/rw/lower_for.cpp
```diff
@@ -0,0 +1,54 @@
...
+                auto if_else_cn = w.cn(w.type_mem());
+                auto if_else    = w.nom_lam(if_else_cn, nullptr);
+                if_else->app(false, brk, {if_else->mem_var(w.dbg("mem")), acc});
+
+                auto cmp    = w.op(ICmp::sl, iter, stop);
+                auto select = w.select(if_then, if_else, cmp);
+                for_lam->app(false, select, mem);
+            }
+
+            return app->rebuild(w, nullptr, {for_lam, app->op(1)}, app->dbg());
```
leissa, line 48: `rebuild` is more supposed to be used, if you want to rebuild a `Def` that you don't know much about. Here you know that you are going to construct an `App`. So better to directly invoke:
```cpp
return w.app(/*...*/);
```

src/thorin/pass/rw/lower_for.cpp
```diff
@@ -0,0 +1,54 @@
...
+const Def* LowerFor::rewrite(const Def* def) {
+    if (auto app = def->isa<App>())
+        if (auto for_ax = app->axiom(); for_ax && for_ax->tag() == Tag::For) {
+            auto& w = def->world();
+            w.DLOG("rewriting for axiom: {} within {}", app, curr_nom());
+
+            auto for_pi  = app->callee_type();
+            auto for_lam = w.nom_lam(for_pi, w.dbg("for"));
+
+            auto body      = app->args()[app->num_args() - 2];
```
leissa, line 16: see comment below. Maybe you want to use a destructering pattern.

src/thorin/pass/rw/lower_for.cpp
fodinabor: neat

src/thorin/pass/rw/lower_for.cpp
fodinabor: I find that too spilly as I only care once for the body and the other args are just all reused as a batch right now.

fodinabor (commented):
Thanks for the feedback!

P.S.: you can also use the "Start a review" button when doing inline comments, then you can submit the whole review as one at the end and one doesn't get an e-mail for each comment ;-)

PR #19, opened 2022-03-17
Closure conversion

f-fries:
Adds typed closure conversion to Thorin.
Test suite can be found  [here](https://github.com/AnyDSL/impala/tree/t2-closure-conv/test/cconv).

closure conversion pass:

-  pass/rw/cconv_prepare: categorize Lam's (returning, firstclass block, free block), eta-expand if necessary
- transform/closure_conv: typed closure conversion
- pass/rw/scalarize: cleanup lambda-lifted functions

closure lowering pass:
- pass/rw/unbox_closures: remove closures for branches
- pass/rw/closure2sjlj: implement first-class continuations with setjmp/longjmp
- pass/fp/closure_analysis: approaximate closure lifetime
- transform/lower_typed_closures: rewrites typed closures to untype closures, firstclass continuations to returning functions

Other Changes:
- CopyProp: Add flag to only propagate through BB-like functions
- EtaRed: Add flag to only apply eta-reduction in callee postion
- BE: generate code for switch
- BE: add support ⊥ return-continuations, gets compiled to unreachable
- BE: add support for first-class functions

**Failing Tests:**
- [x] bench/pidigits.impala: Div by 0 --> Duplicated globals
- [x] bench/nbody: Output mismatch (global planet array is duplicated)
- [x] bench/fannkuch: Output mismatch, global variables
- [x] bench/aobench: Load was considered as FV
- [x] bench/regex: "
- [ ] bench/vcycle: SSA Constr removes retcont + SJ-Code does not behave as expected (?)
- [ ] bench/power_half: Link error


leissa:
Right off the bat some general remarks, before I begin a thorough review:
* doxygen is failing with:

    ```
    lower_typed_closures.h:11: error: end of comment block while expecting command </code> (warning treated as error, aborting now)
    ```
    
* Speaking of doxygen, you can also use markdown to format your doxygen documentation and this is the preferred way of doing it. So you can use backticks instead of `<code> foobar </code>`.

* The naming of "Closure" is a bit inconsistent. Sometimes it's "closure", sometimes it's just "c" as in "CConv". Maybe just use consistently "Clos"? So we have `ClosConv`, `ClosConvPrep`, etc (and Prep is also shorter than Prepare :) )

src/thorin/pass/rw/clos2sjlj.cpp
```diff
@@ -0,0 +1,173 @@
...
+        if (i == 1) { env_seen = true; return env; }
+        return remaining->proj(env_seen ? i - 1 : i - 2);
+    });
+    return (remaining->level() == Sort::Term) ? w.tuple(new_ops) : w.sigma(new_ops);
+}
+
+Lam* Clos2SJLJ::get_throw(const Def* dom) {
+    auto& w = world();
+    auto [p, inserted] = dom2throw_.emplace(dom, nullptr);
+    auto& tlam = p->second;
```
leissa, line 79: double-check that no rehash occurs inbetween. Otherwise, bad things will happen, if you modify the content pointed to by `tlam`.

src/thorin/pass/rw/clos2sjlj.h
```diff
@@ -0,0 +1,42 @@
+
```
leissa, line 1: remove empty line

src/thorin/pass/rw/clos_conv_prep.h
```diff
@@ -0,0 +1,52 @@
+
```
leissa, line 1: remove empty line

src/thorin/transform/clos_conv.h
```diff
@@ -0,0 +1,279 @@
...
+
+std::tuple<Lam*, const Def*, const Def*> clos_lam_stub(const Def* env_type, const Def* dom, const Def* dbg = {});
+
+
+/// @name closure environments
+/// @p tup_or_sig should generally be a @p Tuple, @p Sigma or @p Var.
+/// @{
+
+/// Describes where the environment is placed in the argument list.
+const size_t CLOS_ENV_PARAM = 1_u64;
```
leissa, line 249: -> `Clos_Env_Param` according to coding conventions

leissa (commented):
Most of the stuff looks good to me. There are a few things that look a bit awkward, though :)
Since there are still some failing tests, I will look into these issues more thoroughly later on.

A few more general remarks:
* There is no need for `@brief`. The first sentence will automatically be used as "brief description". So just remove this tag.
* Only use `@p foo`, if `foo` really is a parameter of the function, you are currently documenting. If you just reference another class or method use `this is reference to MyClass` (no special tag needed) or `this is a reference to Myclass::foo` (prefix the method with its class; you even need to do that when cross-referencing methods within `Myclass` itself). Yes, this used to be incorrect in the code base some time ago as well but has been fixed in the meantime. You've probably just oriented yourself there Oo
* Whole sentences should start with a capital letter and end with a full-stop. Make sure that your doxygen doc is made of whole sentences (usually one per line). If you just have really small group names like `getter` then small cap is fine.
* You can easily run [doxygen locally](https://anydsl.github.io/thorin2/index.html#autotoc_md3) to see whether you got the intended result.
* run [`clang-format`](https://anydsl.github.io/thorin2/md__github_workspace_docs_coding.html#autotoc_md27) over your files. There are a couple of things that are not formatted properly.

src/thorin/be/ll/ll.cpp
```diff
@@ -330,6 +352,14 @@ void CodeGen::emit_epilogue(Lam* lam) {
                 args.emplace_back(convert(arg->type()) + " " + emitted_arg);
         }
 
+        if (app->args().back()->isa<Bot>()) {
+            // TODO: Perhaps it'd be better to simply η-wrap this prior to the BE...
+            assert(convert_ret_pi(app->callee_type()->ret_pi()) == "void");
+            bb.tail("call void {}({, })", emmited_callee, args);
+            return bb.tail("unreachable");
+        }
```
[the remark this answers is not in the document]
f-fries: Yes. I think it would be nice to extend RetWap to ensure that all branches and return-continuations are `Lam`s.
I think that's essentially on what's happening on master right now, since after PE only `Lam`s and return-continuations should be left.

src/thorin/pass/fp/eta_red.h
```diff
@@ -9,8 +9,8 @@ namespace thorin {
 /// Rewrites `λx.e x` to `e`, whenever `x` does (optimistically) not appear free in `e`.
 class EtaRed : public FPPass<EtaRed> {
 public:
-    EtaRed(PassMan& man)
-        : FPPass(man, "eta_red") {}
+    EtaRed(PassMan& man, bool callee_only = false)
+        : FPPass(man, "eta_red"), callee_only_(callee_only) {}
```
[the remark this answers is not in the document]
f-fries: Essentially the closure-destruction phase assumes that code-part of a closure is always a Lambda. This flag is used to prevent EtaRed from reducing inside closures.

src/thorin/pass/fp/copy_prop.cpp
```diff
@@ -63,8 +63,8 @@ const Def* CopyProp::rewrite(const Def* def) {
         prop_lam      = var_lam->stub(world(), new_pi, var_lam->dbg());
 
         world().DLOG("new prop_lam: {}", prop_lam);
-        beta_red_->keep(prop_lam);
-        eta_exp_->new2old(prop_lam, var_lam);
+        if (beta_red_) beta_red_->keep(prop_lam);
+        if (eta_exp_)  eta_exp_->new2old(prop_lam, var_lam);
 
         size_t j = 0;
```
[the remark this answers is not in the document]
f-fries: Off the of my head it should work with EtaExp in that context. I'm also not sure why we don't what BetaRed here.

leissa:
So we did a huge rework in `master`. I ported everything but I have to double-check with @fodinabor about the best way how to actually run the closure-conv-related passes. Since I don't want to wait any longer to merge, I switched to the original `optimize` pipeline from `master`. This allows us to merge your work into master and then we can go ahead from there :)

PR #30, opened 2022-04-06
Feature/split splots

leissa:
still WIP


leissa:
post-poned till later

PR #31, opened 2022-04-06
Add some backend-style dot emitter.

fodinabor:
This adds back an option to view the thorin graph. Note, the nodes are clustered into lams following the emitter's schedule.

As an example, the output from the `ForSteps/ForAxiomTest.for_dynamic_iters/17` test after the opt phase:
graphviz

While this is somewhat crude and the text inside the nodes maybe isn't ideal as well (basically `def->dump(0)`), it should help with debugging then and when. The nodes' content can be customized using an optional function parameter for `dot::emit`.

### DIFF
--- a/thorin/CMakeLists.txt
+++ b/thorin/CMakeLists.txt
@@ -38,6 +38,8 @@ add_library(libthorin
     analyses/scope.cpp
     analyses/scope.h
     be/emitter.h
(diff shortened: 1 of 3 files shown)


leissa:
Not that I understand much looking at the graph, but it looks pretty damn cool :D

PR #40, opened 2022-05-17
Autodiff rewrite pass

NeuralCoder3:
A version of automatic differentiation.
In this version (without the sigma change) many features are not supported.

Future improvements:
* working arrays (needs closure conversion)
* better readability
* more modular code design
* better code structure 
* more comments
* consistent formatting of the code
* more unit tests (needs front end)
* solved issues with sigma <-> array conversion


leissa:
Yes, annoying but you have to do it like this on some systems:
```c++
int status = std::system("./test_ad");
EXPECT_EQ(0, WEXITSTATUS(status));
```
to fight this error:
```
ad_test.cpp:169:40: error: lvalue required as unary '&' operand
```
```c++
assert(0 == WEXITSTATUS(std::system("./test_ad")));
```

leissa:
try with my news commit on master that should fix windows issues :)

thorin/world.h
```diff
@@ -145,10 +150,23 @@ class World {
...
+        return lam;
+    }
+
     const Lam* lam(const Pi* pi, const Def* filter, const Def* body, const Def* dbg) {
         return unify<Lam>(2, pi, filter, body, dbg);
     }
     const Lam* lam(const Pi* pi, const Def* body, const Def* dbg) { return lam(pi, lit_true(), body, dbg); }
+    const Lam* flatten_lam(Lam* lam);
+    const Lam* unflatten_lam(Lam* lam);
     ///@}
```
leissa, line 170: I think we should move these helper functions to somewhere else. `World` should just have factory methods (smart constructors). We can either move that to `tuple.*` which already contains some flatten/unflatten helper functions or to `lam.*`.

thorin/pass/rw/auto_diff.h
```diff
@@ -0,0 +1,112 @@
...
+        (this is in agreement with the axiom)
+    and the correct pullback is applied afterwards using the chain rule
+    in fact, returning functions are translated using the axiom
+
+
+Read-only link to overview
+    https://www.overleaf.com/read/gdpfxvzqpfjf
+
+*/
+
```
leissa, line 100: Can you reformat the docs with markdown-style Doxygen? Also, use this comment style:
```c++
/// This is one sentence.
/// Here is another sentence in Doxygen.
/// And here is a third sentence.
/// ```
/// let x = this->is(a, code, example);
/// x.cool_stuff();
/// ```
/// [Draft on Overleaf](https://www.overleaf.com/read/gdpfxvzqpfjf)
/// etc.
```

leissa (changes requested):
Most of the code looks pretty good to me. Great work :+1: 

Besides the more detailed comments below two more general remarks:
* Can you format the code with clang-format? There are couple of odd things like missing spaces between braces etc.
* There are a couple of places with larger blocks of documented code. What's the deal here? Can we remove it or is it still needed?

Once these issues are addressed, I'll look again more thoroughly :)

PR #41, opened 2022-05-19
Rework dialect pass plugin registration & new axiom tagging

fodinabor:
Dialects should implement `thorin_get_dialect_info` to provide a `DialectInfo` that exposes the necessary info and function ptrs for working with it.
`Dialect` is the driver side wrapper around the handle / `DialectInfo`.
The PipelineBuilder is used to compose the various optimization phases in multiple `PassMan`s.

Additionally, as required for some further testing / moving forward with the lowering of dialect axioms, I reworked the identification of axioms.
It now mangles the dialect name into 6 bytes, uses 1 byte to store the index of the axiom of a dialect (group, e.g. *Wrap*) and 1 byte for the flag (Wrap::*add*). This 64bit int is stored in the `fields` of the axiom.
Additionally, the header emission is extended to write out the first 7 bytes as the `dialect::Tag` enum values and provide the infrastructure for matching the dialects' axioms using `dialect::isa<dialect::Tag::group>(def)` and `dialect::isa<dialect::Tag::group>(dialect::group::flag, def)`.
This currently duplicates some code inside the generated headers that can likely be simplified away when the general (and not only parsed) approach for axioms uses all 64bits for the axiom identification.

Todo:
- support dependencies between passes.
- support creating/defining normalizers for axioms from dialects.
- maybe support adding additional phases (`PassMan`s) to the pipeline from the plugin to allow doing orthogonal lowering phases..

Ideas regarding dependencies:
- only support depending on built-in passes, register those first with the `PassMan` and allow querying them from the `PassMan`,
- else, if relying on other dialects is allowed
  - rely on the order of dialect arguments and therefore the registration of plugins with the pass manager and query dependencies from the `PassMan`.
  - register all passes with the `PipelineBuilder` and allow getting dependencies through it.

Additional phases:
- 4 extension points: before opt, after opt, before codegen prep, after codegen prep
- each just a vector of `PassMan`s - caution: possible indeterminism?

Ideas regarding normalizers:
- rely on named resolution, i.e. have `.ax :std.icmp .. , normalize_icmp;` and then in the parser use the `Dialect` to query for the symbol `normalize_icmp`.
- since we bootstrap the dialects anyways, we can add a map from axiom `Tag` to function ptr to the `DialectInfo` that can be used to get the corresponding normalizer function in the parser.


leissa:
Looks already promising :)

I was wondering whether we can simplify this whole `Query`, `Tag2Def`, ... trickery. Since we are generating the header file anyway, we probably need less template magic. I will try to come up with a mock up tomorrow or so.

PR #48, opened 2022-06-08
Extract the mem dialect & allow normalizers in plugins + core normalizers as case study

fodinabor:
First of all, I want to apologize for this monstrosity of a PR... touching some stuff required touching more and more stuff...

The state with this PR is:
- all mem-related axioms are solely available through the dialect and are no longer part of the main Thorin library.
- the LL backend thus had to be moved to the mem dialect plugin.
- plugins can just add passes that they need to `PassMan`, if it's already present, they get a reference to the already present pass. This allows specifying dependencies (`BetaRed`->`CopyProp`,..)
- normalizers can now be defined in the dialect plugins (when specifying a normalizer inside the dialect description e.g. `, normalizer_lea`, the signature is automatically generated and added to a generated `register_normalizers` function, that can be used by the dialect plugin. this also implies that normalizers should only be added to axioms, if the normalizer is actually present)
- as the mem normalizations are not that extensive, I also added a `core` plugin, that contains the normalizations for the axioms from `core.thorin`. A major difference, between the current state of core and mem is, that all `core` axioms are also still defined and supported from the main lib.
- div is in mem, which isn't great..
- adds a big bunch of lit tests that actually run
- supports impala at https://github.com/fodinabor/impala/tree/feature/thorin2-mem-dialect
- ..

### DIFF
--- a/dialects/core/normalizers.cpp
+++ b/dialects/core/normalizers.cpp
@@ -107,10 +107,10 @@ struct Fold<wrap, wrap::shl, w> {
 };
 
 // clang-format off
- template<nat_t w> struct Fold<div, div::sdiv, w> { static Res run(u64 a, u64 b) { using T = w2s<w>; T r = get<T>(b); if (r == 0) return {}; return T(get<T>(a) / r); } };
- template<nat_t w> struct Fold<div, div::udiv, w> { static Res run(u64 a, u64 b) { using T = w2u<w>; T r = get<T>(b); if (r == 0) return {}; return T(get<T>(a) / r); } };
- template<nat_t w> struct Fold<div, div::srem, w> { static Res run(u64 a, u64 b) { using T = w2s<w>; T r = get<T>(b); if (r == 0) return {}; return T(get<T>(a) % r); } };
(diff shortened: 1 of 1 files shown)


leissa:
I just added a bit of cmake magic. what is a bit messy write now is the targets and which targets exactly we need to build for gtests etc. I will look at this tomorrow.

fodinabor:
> Match::sub is now again Match::flags. I guess the background is to still support the legacy hardcoded Axioms with this axiom_from_dialect magic?

Yeah, the new `enum class icmp` elements' values are the full axiom id, so for easy comparison, it makes sense to be able to compare that directly. The `axiom_from_dialect` concept helps to get the return type fitting for `sub` and `flags`: enum vs. int ...

> I just added a bit of cmake magic. what is a bit messy write now is the targets and which targets exactly we need to build for gtests etc. I will look at this tomorrow.

The current state looks good to me.

leissa:
Ah, here is another thing that is a bit odd: Rn, now we have this:
```
./dialects/mem/mem.h
./build/dialects/mem.h
```
which results in confusing includes^^
We could add a suffix to the generated header an place it a bit differently in the build dir like this:
```
./build/dialects/mem/mem_constants.h"
```
Then
```
./dialects/mem/mem.h
```
would just `#include "dialects/mem/mem_constants.h"
```

leissa:
So, should we just merge the `dialects` and `thorin` src trees?

fodinabor:
I'd be in favour of that, yes.

PR #49, opened 2022-06-22
Make thorin output parsable by frontend.

fodinabor:
This should make most thorin output (that only uses known axioms) parsable by the frontend.
Very likely still a bunch of cases not correct...

To simplify outputting nom var names, the parser now supports reading a list of var names after the nom's type.
The proposed syntax (to disambiguate from ops) is `.lam name : type, @(var0, var1) = {..};`

Known issues:
- nominal sigmas are emitted as `.Sigma name : type, #ops, @(var,..) = { [..] };`, which cannot be parsed at the moment. To support this, we either need integrated special cases in `parse_nom` and `parse_def` (similar to `parse_sigma`), or great infer support (where even the types of the infer nodes are unknown until parsing the definition..)

### DIFF
--- a/thorin/world.h
+++ b/thorin/world.h
@@ -572,7 +572,6 @@ public:
         swap(w1.err_,      w2.err_);
         // clang-format on
 
-        swap(w1.data_.imported_dialects_, w2.data_.imported_dialects_);
         swap(w1.data_.univ_->world_, w2.data_.univ_->world_);
         assert(&w1.univ()->world() == &w1);
         assert(&w2.univ()->world() == &w2);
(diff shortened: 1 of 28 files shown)


thorin/world.h
```diff
@@ -568,6 +572,7 @@ class World {
         swap(w1.err_,      w2.err_);
         // clang-format on
 
+        swap(w1.data_.imported_dialects_, w2.data_.imported_dialects_);
```
leissa, line 575: shouldn't this be covered by the
```c++
swap(w1.data_,     w2.data_);
```
above?

fodinabor: should be fixed. The field should be copied upon `World::stub`.

PR #50, opened 2022-06-23
Add `make install` support.

fodinabor:
I did not change the directory layout of thorin for now. Some things I'd preferably change, like the fact that we pollute the `include` folder with `dialects`. Also, we could remove some matching overhead if we had a single `include` folder, but it's not an incredible overhead.

This installs thorin to `CMAKE_INSTALL_PREFIX` with the following structure:
```
bin
\ thorin
include
\ absl
\ dialects
\ thorin
\ half.hpp
lib
\ cmake
  \ absl
    \ *.cmake
  \ thorin
    \ thorin-config.cmake
    \ thorin-config-version.cmake
    \ thorin-targets-[debug|release].cmake
    \ thorin-targets.cmake
    \ Thorin.cmake
\ thorin
  \ *.thorin
  \ libthorin_*.so
\ libabsl_*.so
\ libthorin.so
```

After installing, external dialects just need to find the package:
```
cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
project(dialect)

find_package(thorin)

add_thorin_dialect(dialect
    SOURCES
        dialect/dialect.h
        dialect/dialect.cpp
)
```
and use `cmake .. -Dthorin_DIR=<THORIN_INSTALL_PREFIX>/lib/cmake/thorin` to configure the project.

Other projects using thorin internally, such as Impala, could use `find_package(thorin)` and then link against `thorin::libthorin`.

One possible point of discussion is whether setting `set(ABSL_ENABLE_INSTALL ON)` really is a good idea, or whether we should instead list all targets that we transitively depend on manually, like so: `install(TARGETS absl_.. absl_.. EXPORT absl_export)`. With the current solution, Abseil complains, when `CMAKE_INSTALL_PREFIX` is not set to some local folder: `The default and system-level install directories are unsupported except in LTS   releases of Abseil.  Please set CMAKE_INSTALL_PREFIX to install Abseil in your   source or build tree directly.` Another option to silence that, is likely switching to a tagged version of Abseil?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,13 +48,14 @@ include(CTest)
 if(BUILD_TESTING AND THORIN_BUILD_TESTING)
     include(GoogleTest)
     set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-    add_subdirectory(modules/googletest)
+    add_subdirectory(modules/googletest EXCLUDE_FROM_ALL)
 endif()
 
 set(ABSL_PROPAGATE_CXX_STD ON)
 set(ABSL_USE_EXTERNAL_GOOGLETEST ON)
+set(ABSL_ENABLE_INSTALL ON)
 add_subdirectory(modules/abseil-cpp)
-add_subdirectory(modules/lyra)
+add_subdirectory(modules/lyra EXCLUDE_FROM_ALL)
 
 add_subdirectory(thorin)
 add_subdirectory(cli)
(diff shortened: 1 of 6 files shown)


leissa:
Hmm, we actually are on a tagged version of absl: `lts_2021_11_02`. So I don't really understand the warning. With
```cmake
install(TARGETS absl_.. absl_.. EXPORT absl_export)
```
we would only install the parts of absl that we really need, correct?

fodinabor:
Apparently we're not. We are on [ca80034](https://github.com/abseil/abseil-cpp/tree/ca80034f59f43eddb6c4c72314572c0e212bf98f) which apparently is some random commit on master. When checking out the mentioned lts branch locally, it silences the warning.

Yes, the manual install specification should allow us, to limit building and installing to the parts of abseil that we need.
By doing so, we'll be installing an incomplete version of abseil, though. In general, just installing abseil along with installing thorin feels a bit intrusive (conflicts with other installations of abseil, when installing to global locations like `/usr`?), but I'm not really certain what other options we have..

leissa:
ah, yes, I misread my gitk output. I think I just used `master` when I added the repo. I think there was a problem with the latest official release - sth with mixing gtest and absl didn't went well with cmake. Sth like this. I don't think we really have another option than bundling thorin with absl. 3rd-party dialects really require to be built with exactly **our** version of absl. We could add an option to use the system install of absl instead of the submodule. But I think for now, the current solution works. Let's just ignore the warning and bump to the next lts version.

PR #51, opened 2022-06-24
Add %core.conv and %core.bitcast axioms

fodinabor:
I need those axioms for some array-related tests..

This includes allowing to specify axioms where the subs have differing types.

I also introduced a new dependency type for dialects: if the plugin just needs the generated header of another dialect, this can be specified separately now. This e.g. allows using `core::conv` inside the `clos` and `mem` dialects without creating actual circular dependencies.

Impala version: https://github.com/fodinabor/impala/tree/feature/cast-in-dialect


leissa:
Fantastic work again :+1:

PR #53, opened 2022-06-28
Feature/output

leissa:


leissa:
You can specify a prefix and override the output file for each emitter like so:
```
./thorin bla.thorin -e ll -e h -e md -o foo ---output-ll bla.ll
```
This will generate:
* `bar.ll`
* `foo.h`
* `foo.md`

In addition you can specify `-` as output file to dump to cout:
```sh
./thorin bla.thorin -e thorin --output-thorin -
```

This feature is used to add an `_autogen.h` suffix to the auto-generated dialect headers.

PR #56, opened 2022-06-30
Feature/ptrn

leissa:
This introduces patterns to Thorin's frontend. Originally, I intended not to introduce an AST layer but for patterns it makes sense as they don't duplicate anything that is already present within Thorin. Also, the Ptrn AST nodes are really lightweight.

Rn, it is only used within let expressions like so:
```
.let (x, y) = my_tuple;
```
You can also optionally annotate the type of an IdPtrn:
```
.let (x: T, y) = my_tuple;
```
But `T` is not used for anything right now. We can either use that later on to help type inference or we can remove this again.

There are also these var_lists which I believe we can remove in favor of proper patterns. And we somehow have to come up with a syntax that kind of blends patterns with dependent Sigmas.

### DIFF
--- a/docs/langref.md
+++ b/docs/langref.md
@@ -126,7 +126,7 @@ The following tables comprise all production rules:
 | Nonterminal | Right-Hand Side                                                   | New Scope? | Comment                        | Thorin Class  |
 |-------------|-------------------------------------------------------------------|------------|--------------------------------|---------------|
 | d           | `.ax` Ax `:` e<sub>type</sub> `;`                                 |            | axiom                          | thorin::Axiom |
-| d           | `.let` Sym `:` e<sub>type</sub> `=` e `;`                         |            | let                            | -             |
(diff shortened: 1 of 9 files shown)


fodinabor (approved):
Looks good to me :)

so we can now basically use `.let (a, b, c) = @nom_id;` as a replacement for the `var_list`?

Just a note: the rules for the var reference (`@nom_id`) are missing from the docs.

leissa:
> so we can now basically use `.let (a, b, c) = @nom_id;` as a replacement for the `var_list`?

yes. I was thinking in having sth even better like this:

```
.lam (x: i32, y: i32) -> i32 { /*...*/ }
```
But I'm a bit unsure how this all fits together with dependent Sigmas. We could end up with sth like this:

```

.lam (T: *, x: f T) -> g T { /*...*/ }

```

PR #57, opened 2022-07-01
bump abseil and googletest versions

leissa:
From abseil release notes:

> CMake builds now use the flag ABSL_BUILD_TESTING (default: OFF) to
> control whether or not unit tests are built.

So, BUILD_TESTING/BUILD_TESTING_BACKUP trickery not necessary anymore :)


leissa:
This also fixes this abseil LTS warning.

PR #59, opened 2022-07-04
Direct style to continuation-passing style

NeuralCoder3:
This dialect provides a conversion from direct style to continuation-passing style.

Idea:
Convert each direct style call into the body of the calling lambda and create a new lambda accepting the result as an argument.
The new lambda gets the old body where the call is replaced by the argument.

```
h:
  b = f a
  C[b]
```
becomes
```
h:
    f'(a,h_cont)

h_cont(b):
    C[b]
```
with the following types:
```
f : A -> B
f': .Cn [A, ret: .Cn[B]]
```

Available axioms and passes:
* `ds2cps : (T -> U) -> Cn [T, Cn U]` (currently not implemented)
* `cps2ds : Cn [T, Cn U] -> (T -> U)`
* `ds2cps` pass: converts direct style calls and involved functions to cps as describes above

`cps2ds` generates no code. It is instead resolved in the `ds2cps` pass at direct style caller sites.

Applications:
* easy function definitions from the point of view of somebody using the thorin frontend
* easy rewriting of direct-style axioms with direct style functions or wrapped cps functions

Note: 
The name of the dialect is `direct` due to the eight-character limit that prevents more informative names.


dialects/direct/direct.thorin
```diff
@@ -0,0 +1,26 @@
...
+///
+/// ## Operations
+///
+/// ### %direct.ds2cps
+///
+.ax %direct.ds2cps: Π [T: *, U: *] -> (T -> U) -> .Cn [T, .Cn U];
+///
+/// ### %direct.cps2ds
+///
+.ax %direct.cps2ds: Π [T: *, U: *] -> (.Cn [T, .Cn U]) -> (T -> U);
```
leissa, line 26: add newline

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
+#include "dialects/direct/passes/ds2cps.h"
+
+#include <iostream>
+
+#include <thorin/lam.h>
+
+#include "dialects/direct/direct.h"
+
+// #define verbose_rewrite
```
leissa, line 9: plz remove

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
...
+    Lam* lam = curr_nom();
+    rewrite_lam(lam);
+}
+
+bool isDS(Lam* lam) {
+    auto ty       = lam->type()->as<Pi>();
+    auto codom_ty = ty->codom();
+    return !(codom_ty->isa<Bot>());
+}
+
```
leissa, line 26: There is already a getter `Pi::is_cn`. Also, note that `Lam::type` overrides `Def::type` and already returns `Pi`. So you can remove this one here and simply use it like this:
```c++
isDS(lam)
```
->
```c++
!lam->type()->is_cn()
```

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
...
+#ifdef verbose_rewrite
+    std::cout << "DS2CPS: " << lam->name() << " : " << ty << " = " << arg_ty << " => " << ret_ty << std::endl;
+#endif
+
+    // overwrite lam body (or new lambda)
+    auto result = rewrite_(currentLambda->body());
+#ifdef verbose_rewrite
+    std::cout << "set body of " << currentLambda->name() << " to " << result << " : " << result->type() << std::endl;
+#endif
+    currentLambda->set_body(result);
```
leissa, line 50: Can you remove these `ifdef`s in favor of `world().DLOG(/*...*/)`?

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
...
+/// replaces the body
+/// note: if the lambda contains a ds call, the body will be replaced
+///   with a cps call to the previously ds function and the original
+///   body will be altered and stored in a new continuation
+void DS2CPS::rewrite_lam(Lam* lam) {
+    if (auto i = rewritten_.find(lam); i != rewritten_.end()) return;
+
+    Lam* prev     = currentLambda;
+    currentLambda = lam;
+
```
leissa, line 37: just for the sake of consistency:
`currentLambda` -> `curr_lam` or maybe just `curr`. It's also just `prev` so I guess `curr` is just fine :)

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
...
+/// rewrites ds calls to cps calls
+/// descents through all other definitions
+const Def* DS2CPS::rewrite_inner(const Def* def) {
+    World& world = def->world();
+
+#ifdef verbose_rewrite
+    std::cout << "rewrite " << def << " : " << def->type() << std::endl;
+#endif
+
```
leissa, line 83: `world().DLOG` see above

dialects/direct/passes/ds2cps.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    return def;
+}
+
+PassTag* DS2CPS::ID() {
+    static PassTag Key;
+    return &Key;
+}
+
```
leissa, line 265: can you also format with `clang-format`? There are some odd formattings here and there.

leissa (commented):
Great work :+1:

fodinabor (commented):
Looks good in general.
From my side just two CMake comments + running `clang-format` would be nice :)

dialects/direct/passes/ds2cps.cpp
NeuralCoder3: I forgot. 
:+1: 
I should probably enable format on save.

PR #60, opened 2022-07-04
Feature/pattern

leissa:
Adds a binder syntax mirroring the pattern syntax.
This is still WIP.


fodinabor (approved):
I guess it looks good enough to me, that we should merge before it get's any bigger... :)

PR #61, opened 2022-07-04
Hopefully fix parallel make builds.

fodinabor:
See 4 & 5 in https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/


leissa:
Seems to work on my notebook :)

PR #62, opened 2022-07-15
Make DS2CPS more robust

fodinabor:
DS2CPS both currently either bail or fail if a loop matches their patterns.

DS2CPS just died of a stack overflow due to infinitely following the loop recursion.
DS2CPS should generally be a bit more stable now.
One major downside of this change is, that when having a loop, we must not set the PE filter to true, but without doing so, we introduce some unnecessary overhead for the non-loop cases.
Note, the loop support greatly depends on copyprop being able to propagate the continuations through branches, which I removed from this patch.

### DIFF
--- a/dialects/direct/passes/ds2cps.cpp
+++ b/dialects/direct/passes/ds2cps.cpp
@@ -25,7 +25,7 @@ void DS2CPS::rewrite_lam(Lam* lam) {
     if (auto i = rewritten_.find(lam); i != rewritten_.end()) return;
 
     Lam* prev = curr_lam_;
-    curr_lam_  = lam;
+    curr_lam_ = lam;
 
     auto ty     = lam->type()->as<Pi>();
     auto arg_ty = ty->dom();
@@ -55,6 +55,8 @@ const Def* DS2CPS::rewrite_(const Def* def) {
     if (auto i = rewritten_.find(def); i != rewritten_.end()) return i->second;
 
     if (auto lam = def->isa_nom<Lam>()) {
+        if (!isa_workable(lam)) return lam;
+
         auto ty = lam->type();
         if (!ty->is_cn()) {
             // extend ds function with return continuation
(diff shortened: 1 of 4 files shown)


leissa:
This looks overly complicated to me. 

> CopyProp needed to be extended to explicitly work on branches `(f, t)#cond (args, ...)` instead of just directly applied non lams.

IMHO only dealing with directly applied nom lams is enough due to the EtaExp - which makes this whole `CopyProp::rewrite_conditional` stuff superfluous. Maybe there is another problem with the EtaExp or its interaction with the CopyProp.

leissa (changes requested):
The DS2CPS changes look pretty solid to me, but I had a concern regarding the CopyProp (see above).

fodinabor:
The `lit/main_loop_cn.thorin` test case fails for master, but is supported by this branch. The main issue here is that the continuations have to be fully removed for the ll backend to be happy. Afaict, EtaExp would not get rid of those parameters. LamSpec does not support this either and would be complicated to adapt to it (as an RWPass) and for the cases that I came up with, CopyProp seemed to be able to solve these situations by propagating through the branch as well.

I'm currently still investigating an issue with the CopyProp, though (calling the loop from the test case twice - once from main, once from the exit continuation).

fodinabor:
I removed the CopyProp changes from this PR. I currently don't see a straightforward way to avoid that DS2CPS generates a loop where all lams have a continuation. The input for the CopyProp was already Eta reduced, which made it pretty clean.

I've been considering adding an extra PM phase so we can use the CopyProp changes without it fighting against the Beta reduction, as it currently does for some cases, but that would again require tighter coupling from the `optimize` function to the plugins since CopyProp is in the mem dialect. Before I do that, I'm wondering whether we might want to generalize the pass registration further:
I could imagine that it'd make sense to have a map similar to the backends map that is currently used for the `ll` backend. This map is filled by the plugins with `pass->name()` -> `<Pass>* thorin_add_<dialect>_<pass>(PassMan& pass)`.
This has two benefits:
- we can write textual descriptions of the pass pipeline, we want to use (think LLVM's new `opt --passes="function(<pass>,<pass>)"` syntax, where `()` can be used to group passes together into a PM.
- we can more easily build up a custom pipeline in frontend X, since we just load all plugins we need, and don't have to ask `dlsym` for the creation functions, but instead can query the map directly with the passes' name.

Thoughts?

leissa:
I'll merge for now. We can discuss your other points next time in our online meeting :)

PR #63, opened 2022-07-21
Cleanup/world state

leissa:
* some cleanup
* makes breakpoints available during the construction of the world
* makes various hash containers deterministic again

### DIFF
--- a/thorin/world.cpp
+++ b/thorin/world.cpp
@@ -32,8 +32,9 @@ namespace thorin {
 bool World::Arena::Lock::guard_ = false;
 #endif
 
-World::World(std::string_view name)
-    : checker_(std::make_unique<Checker>(*this))
+World::World(std::string_view name, const State& state)
+    : state_(state)
+    , checker_(std::make_unique<Checker>(*this))
     , err_(std::make_unique<ErrorHandler>()) {
     data_.name_        = name.empty() ? "module" : name;
     data_.univ_        = insert<Univ>(0, *this);
@@ -195,19 +196,13 @@ World::World(std::string_view name)
     }
 }
 
+
+World::World(std::string_view name)
+    : World(name, State())
+{}
+
 World::~World() {
     for (auto def : data_.defs_) def->~Def();
-}
-
-World World::stub() {
-    World w(name());
-    w.state_ = state_;
-
-    // bring dialects' axioms into new world.
-    Rewriter rewriter{w};
-    for (const auto& ax : data_.axioms_) rewriter.rewrite(ax.second);
-
-    return w;
 }
 
 /*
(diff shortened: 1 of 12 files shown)


thorin/world.cpp
```diff
@@ -199,19 +200,6 @@ World::~World() {
...
 
-World World::stub() {
-    World w(name());
-    w.ostream_                 = ostream_;
-    w.state_                   = state_;
-    w.data_.imported_dialects_ = data_.imported_dialects_;
-
-    // bring dialects' axioms into new world.
-    Rewriter rewriter{w};
-    for (const auto& ax : data_.axioms_) rewriter.rewrite(ax.second);
```
fodinabor, line 210: Is this removed without replacement?
I believe that is quite dangerous:
- `%mem.malloc` is parsed early on and added to the initial world, referencing types / params from the original world
- `%mem.alloc` is used in the actual program (reachable from a `.lam .extern`
- `%mem.alloc`->`%mem.malloc` is rewritten by `alloc2malloc`, or so it tries, but since the `%mem.malloc` still references the original world it is quite flakey

fodinabor (changes requested):

thorin/world.cpp
leissa: This kind of flew under my radar and I was debugging for an hour or so before I realized that I removed these two crucial lines XD
Anyway, I have now added this directly to the Rewriter. If we make axioms external (which makes sense to me), we can remove this altogether.

fodinabor: LGTM.
That should be alright, since `data_` is no longer automatically copied, so at least there are no broken axioms in the new world when using the `World(State &)` ctor 👍🏼

I agree that it'd be convenient to have axioms as externals for this case at least.
Not sure I'd be happy with it though, the name external is kind of misleading. I understand the external things to be accessible externally, e.g. via FFI calls (from C), which is not the case for axioms.

fodinabor (approved):
LGTM now :)

leissa:
Just one thing I stumbled upon:
Rn, there is this `State::imported_dialects` thing. This doesn't really make sense to have this as some kind of set bc we want to maintain the original order, no?

fodinabor:
well, we certainly don't want to have an import stored in there twice, so a set was quite a canonical choice.
And afaict, we don't really care in which order the imports happen.. if something depends on something else, it should just import it themselves..

leissa:
> well, we certainly don't want to have an import stored in there twice, so a set was quite a canonical choice. And afaict, we don't really care in which order the imports happen.. if something depends on something else, it should just import it themselves..

Good point. Then, my use a `absl::btree_set` for this to make things deterministic actually makes sense :)

PR #68, opened 2022-08-04
Refactor/dump

leissa:
Keep improving the output ...


fodinabor (approved):
Looks fine to me.

Though I'm starting to guess, that actually useful (reparsable) output we'll only ever get with a schedule, so that `.let`s are always put into the outer most scope where they are used..

leissa:
There are several bigger issues.
1) as you pointed out, we need to schedule the Defs to put them in a correct block. This is sth we could add to the DepTree.
2) If we have sth like this:
```
.lam f (n: .Nat, x: F n) -> ... = ... n ...
```
Note that `n` means two different things depending on the context: First, it's a Var for the dependent Sigma. Second, it's a var for the Lam.

PR #70, opened 2022-08-07
Feature/phase

leissa:
First impl of the discussed Phase infrastructure. 
* I already subclassed two phases from the clos dialect from `Phase`. I think we could also remove some boilerplate by subclassing them from `RWPhase`.
* Most of `RWPhase` and `FPPhase` is not yet tested. We probably want to have at least one demo phase for each flavor. (But `Cleanup` is a simple subclass of `RWPhase`).
* Ported the dot and llvm backend to `ScopePhase`.
    This also moves the `World::visit` stuff to the phase infrastructure. Having it inside `World` always felt a bit out of place and fights the god class status of `World`.
* `optimize` now just builds a `Pipeline`
* Some cleanups here and there regarding `Pass`/`Phase`


leissa:
Alright, I think I have everything I initially wanted to have in this PR and I'm ready to merge and let me know what you think.

thorin/phase/phase.h
```diff
@@ -0,0 +1,199 @@
...
+        , new_world_(world.state())
+        , rewriter_(*this, world, new_world_) {}
+
+    void start() override;
+
+    /// @name getters
+    ///@{
+    World& old_world() { return world(); }
+    World& new_world() { return new_world_; }
+    Def2Def& old2new() { return rewriter_.old2new; }
```
fodinabor, line 78: So the idea with the RWPhase is that you edit the `old2new` map upon construction of a subclass or override `{pre,post}_rewrite` to do something else than `Cleanup`?

fodinabor (changes requested):
There's one blocker, see comment on `ID` function.

Apart from that, it looks great!
This allows building a custom pipeline from a frontend as usually wanted, nicely integrating "free function" transformations :+1:

The extensibility from dialects currently still is "just" in the two PassMan phases, but as discussed, to do this properly, we'd probably need to extend that using some textual pipeline description and another map from phase names to phases that is filled by the dialects' register functions. This obviously is a task for another day.

thorin/phase/phase.h
leissa: The idea is that you just look at the `old_world` while constructing a `new_world` along the way. Maybe, I should change that to
```c++
const World& old_world() { return world(); }
```
And yes, you would override `pre/post_rewrite`. If you just want to do a simple rewrite, you can just use the `Rewriter` for that. We can also add a `Phase` wrapper for that if this is sth that occurs regularly.

thorin/pass/pass.h
```diff
@@ -240,17 +227,22 @@ class RWPass : public Pass {
         else
             return man().curr_nom()->template as<N>();
     }
+
+    static const PassTag* ID() {
+        static PassTag Key;
+        return &Key;
+    }
```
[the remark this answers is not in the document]
leissa: see my latest commit for a different take

fodinabor (approved):

PR #73, opened 2022-08-28
Curried Function

leissa:
Adds syntactic sugar for defining "inline" curried functions.
* [x] parsing
* [x] actually construct IR
* [x] filter


leissa:
I now have it implemented. You can now do sth like this:
```
.lam id(T: *)(x: T) -> T = x;
```
That being said, there are a couple of issues in other parts in Thorin that prevent this from working. E.g. some optimizations assume that all Lams are continuations; others going bonkers when dealing with dependent stuff. But this is not the issue of the front end, so I suggest to merge and investigate these issues afterwards.

You can also specify filters. By default, the filter is set to `.ff`. You can specify the filter like this:
```
.lam f(x: T, y: U)@(..x..y..)(a: A, b: B)@(..x..y..a..b..) = ...;
```
In addition, since setting the filter to true is such a common use case, you can put a bang **in front** of the pattern:
```
.lam f!(x: T, y: U)(a: A, b: B) = ...;
```
Here, the outer lam receives a `.tt` filter and the inner lam a `.ff` filter.

fodinabor (approved):
Looks good for me so far - not sure whether we want to merge this as long as it's still going completely bonkers though..?
Does this alone solve any pressing issue of @NeuralCoder3 ?

leissa:
Should not break any of the existing stuff. Only if you start using curried functions - and this issue is present in `master` as well. It's just cumbersome to write it down.

NeuralCoder3 (approved):

PR #74, opened 2022-08-30
Refactor/dep

leissa:
@NeuralCoder3 does this work for you?

Note, there is also an impala branch `refactor/dep`.


NeuralCoder3 (approved):
Yes, this solves the non-rewrite issues with the axioms.

PR #75, opened 2022-08-31
Debug Dialect

NeuralCoder3:
A simple dialect that prints debug information of symbols.
The main reason for the dialect are:
* easy inspection of elements that can be invoked in thorin
* minimal demonstration of a plugin with a pass, lit tests, axioms, and normalizers
* template to create new plugins
* orientation point for dialect developers regarding changes in the dialect or pass infrastructure

The dialect provides two axioms:
`dbg: Π [T:*] -> T -> T`
This axiom prints its content and destroys itself afterward in the normalizer.

`dbg_perm: Π [T:*] -> T -> T`
This axiom also prints its content in the normalizer but remains.
It is only removed at the codegen prep phase.
Caution: It is not transparent to other dialects and may alter the program's behavior.


dialects/debug/debug.cpp
```diff
@@ -0,0 +1,38 @@
...
+extern "C" THORIN_EXPORT thorin::DialectInfo thorin_get_dialect_info() {
+    return {"debug",
+            [](thorin::PipelineBuilder& builder) {
+                builder.extend_codegen_prep_phase([](PassMan& man) { man.add<thorin::debug::DebugRemovePerm>(); });
+            },
+            nullptr, [](Normalizers& normalizers) { debug::register_normalizers(normalizers); }};
+}
+
+#define YELLOW "\033[0;33m"
+#define BLANK  "\033[0m"
```
leissa, line 24: I believe this doesn't work on Win, and I think we have some preprocessing macro for this at some other place, but since I don't use Win I don't care so much XD
So, unless you want to look into it, I'd suggest to simply keep it like this.

dialects/debug/passes/remove_perm.h
```diff
@@ -0,0 +1,23 @@
...
+
+/// removes all permanent debug markers for code gen
+class DebugRemovePerm : public RWPass<DebugRemovePerm, Lam> {
+public:
+    DebugRemovePerm(PassMan& man)
+        : RWPass(man, "debug_remove_perm") {}
+
+    const Def* rewrite(const Def*) override;
+
+    static PassTag* ID();
```
leissa, line 16: remove. not needed anymore.

dialects/debug/passes/remove_perm.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <thorin/lam.h>
+#include <thorin/tables.h>
+
+#include "dialects/debug/debug.h"
+
+namespace thorin::debug {
+
+const Def* DebugRemovePerm::rewrite(const Def* def) {
+    auto& world = def->world();
+
```
leissa, line 14: There is a `world()` getter available from `Pass`. So remove this line ...

dialects/debug/passes/remove_perm.cpp
```diff
@@ -0,0 +1,29 @@
...
+    if (auto dbg_perm_app = match<dbg_perm>(def)) {
+        auto e = dbg_perm_app->arg();
+        world.DLOG("dbg_perm: {}", e);
```
leissa, line 17: and use
```c++
world().DLOG ...
```
here

dialects/debug/passes/remove_perm.h
```diff
@@ -0,0 +1,23 @@
...
+
+    const Def* mod_pb(const Def* def);
+
+private:
```
leissa, line 20: dangling `private`. remove.

dialects/CMakeLists.txt
```diff
@@ -72,3 +72,13 @@ add_thorin_dialect(direct
...
+
+add_thorin_dialect(debug
+    SOURCES
+        debug/debug.h
+        debug/debug.cpp
+        debug/passes/remove_perm.h
+        debug/passes/remove_perm.cpp
+        debug/normalizers.cpp
+    INSTALL
+)
```
leissa, line 84: newline.

PR #78, opened 2022-09-14
parse filter

NeuralCoder3:
A simple addition to the parser 
for parsing filters.

A nominal function now first parses 
an expression after `=` and checks if a `,` follows.
If this is the case, the parsed expression is used as filter
and the body is parsed. 
Otherwise, the parsed expression is used as body.

This way (as opposed to moving the filter into the body),
the filter is near the lambda definition and separated from the body.

### DIFF
--- a/thorin/fe/parser.cpp
+++ b/thorin/fe/parser.cpp
@@ -755,8 +755,14 @@ void Parser::parse_nom_fun() {
     dom_p->bind(scopes_, lam_var);
 
     if (accept(Tok::Tag::T_assign)) {
-        auto body = parse_expr("body of a lambda");
-        lam->set(false, body);
+        const Def* filter = world().lit_bool(false);
+        auto body         = parse_expr("filter/body of a lambda");
+        if (ahead().tag() == Tok::Tag::T_comma) {
+            eat(Tok::Tag::T_comma);
+            filter = body;
+            body   = parse_expr("body of a lambda");
+        }
+        lam->set(filter, body);
     }
     expect(Tok::Tag::T_semicolon, "end of lambda");
 


fodinabor:
So we'll have something like the following?
```rust
.cn add [lhs : %Int 256, rhs : %Int 256, cont: .cn [%Int 256]] = .tt, {
   cont (%core.wrap.add (0, 256) (lhs, rhs))
};
```
I guess in general I'm fine with that, though I am wondering whether we want the explicit distinction to the nom array's / sigma's / .. shape parameter or we'd want to do it similar to that, i.e. move the filter before the assignment:
```rust
.cn add [lhs : %Int 256, rhs : %Int 256, cont: .cn [%Int 256]], .tt = {
   cont (%core.wrap.add (0, 256) (lhs, rhs))
};
```

NeuralCoder3:
A filter only makes sense in the presence of a body. Therefore, I would move it after the equal sign to group it together with the body.

leissa:
I mean, we can just add in the parser that if you do sth like this:
```
.lam (...)@(...);
```
That you are not allowed to do that.

NeuralCoder3:
> Anyway, I started working on this curried function where you need a mechanism to specify a filter for each currying level. Sth like this:
> 
> ```
> .lam (T: *, U: *, flag: .Bool)@(.tt)(x: F T, y: G U)@(flag) = ...
> ```

I can also agree with this design.

leissa:
If you urgently need this feature, I suggest the following:
Change the parser in your patch to the proposed syntax by me. Then we merge. And later on I just put #73 on top of that.

PR #83, opened 2022-09-20
Refactor/int

leissa:
This patch replaces the axiom `%Int` with a builtin `.Int`. 

`Int` is so deeply geared into Thorin that it doesn't really make sense to have it as an "optional" axiom. What is more, we sometimes get weird bootstrapping problems because of this.

* FE syntax changes from `%Int` to `.Int` (as already discussed: we cannot really make `%Int` part of a dialect)
* removed `isa_sized_type` in favor of `isa<Tag::Real>(def)` or `def->isa<Int>()`

### DIFF
--- a/dialects/affine/affine.thorin
+++ b/dialects/affine/affine.thorin
@@ -23,7 +23,7 @@
 ///
 /// After termination of the loop `exit` is invoked.
 .ax %affine.For: Π [m: .Nat , n: .Nat , Ts: «n; *»] ->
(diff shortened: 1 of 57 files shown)


leissa:
As a next step, I'd like to rename `.Int` to `.Idx` (and `world.type_int(n)` to `world.type_idx(n)`) and use the terminology `.Int n` = `.Idx 2^n`. The current one is really confusing.

fodinabor (approved):
I suppose it makes sense to me to do this change to emphasize the Idx's builtin nature and looks good as far as I can tell.

NeuralCoder3 (approved):

PR #84, opened 2022-09-21
Refactor/int2idx

leissa:
* `Int` -> `Idx`
* `.Int` -> `.Idx`
* `lit_int_width` -> `lit_int_`
* `type_int_width` -> `type_int_`
* `lit_int` -> `lit_idx`
* `lit_int_mod` -> `lit_idx_mod`
* `type_int` -> `type_idx`

This effectively renames `Int` to `Idx` and uses the term `int w` to refer to an `.Idx 2^w`. Note that I renamed `lit_int_width` to `lit_int_`. This gives you some time to port to the new API and get type errors instead of weird runtime errors.

### DIFF
--- a/dialects/affine/affine.h
+++ b/dialects/affine/affine.h
@@ -9,7 +9,7 @@ namespace thorin::affine {
 /// Returns the affine_for axiom applied with \a params.
 /// See documentation for %affine.For axiom in @ref affine.
 inline const Def* fn_for(World& w, Defs params) {
-    return w.app(w.ax<affine::For>(), {w.lit_nat(width2mod(32)), w.lit_nat(params.size()), w.tuple(params)});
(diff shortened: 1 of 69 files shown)


fodinabor:
Nice :)
What would be options to support int in the fe?
I guess we could have the axiom `%core.Int` whose normalizer just makes it a `.Idx`
Or we can have just a type ctor in the core dialect: `.cn core.Int : w: .Nat -> * = { .Idx (pow(2,w)) };` (this obviously requires having `pow` defined there as well..)
Any alternatives?

leissa:
I think as a next step, I'd like to make this special `%foo.bar.baz` namespace available for any entity. This way, we could do this:
```
.lam %core.Int (w: .Nat)@(.tt) = .Idx pow(2, w);
%core.I8 = %core.Int 8
%core.I16 = %core.Int 16
%core.I32 = %core.Int 32
%core.I64 = %core.Int 64
```

PR #85, opened 2022-09-24
Refactor/porting to core

leissa:
Ports the remaining axioms to dialects.


leissa:
Since we don't have two competing matching systems anymore, I also cleaned up this part.

leissa:
[Impala](https://github.com/anyDSL/impala/tree/refactor/porting_to_core) also works.

thorin/world.h
```diff
@@ -711,27 +583,12 @@ class World {
...
-        std::array<const Axiom*, Num<PE   >> PE_;
-        std::array<const Axiom*, Num<Acc  >> Acc_;
-        // clang-format on
         const Lit* lit_nat_0_;
         const Lit* lit_nat_1_;
         const Lit* lit_nat_max_;
         const Lit* lit_univ_0_;
         const Lit* lit_univ_1_;
-        const Axiom* atomic_;
         const Axiom* bitcast_;
```
fodinabor, line 591: the initialization code is removed, do we still need this?

dialects/core/core.thorin
```diff
@@ -122,26 +54,166 @@
...
 /// Bitcast to reinterpret a value as another type.
 /// Can be used for pointer / integer conversions as well as integer / nat conversions.
 .ax %core.bitcast: Π [D: *, S: *] -> S -> D, normalize_bitcast;
+///
+/// ## Other Operations
+/// 
+/// ### %core.trait
+/// 
+/// Yields the size or align of a type.
+.ax %core.trait(size, align): * -> .Nat, normalize_trait;
```
fodinabor, line 206: Just a random note... how do we implement traits for other dialects' axioms?

dialects/mem/mem.cpp
```diff
@@ -22,17 +23,33 @@ using namespace thorin;
...
+                    man.add<mem::RememElim>();
+                    man.add<mem::Alloc2Malloc>();
+                });
             },
             nullptr, [](Normalizers& normalizers) { mem::register_normalizers(normalizers); }};
 }
+
+namespace thorin::mem {
+
+const Def* op_malloc(const Def* type, const Def* mem, const Def* dbg) {
```
fodinabor, line 43: I understand the urge for non-inline constructors, but moving this here means we have to export this function explicitly and other dialects have to link against the mem dialect plugin if we want them to be able to use this...

thorin/axiom.h
```diff
@@ -57,104 +59,67 @@ class Axiom : public Def {
...
     const D* operator->() const { return def_; }
     operator const D*() const { return def_; }
     explicit operator bool() { return axiom_ != nullptr; }
+    ///@}
 
-private:
-    const Axiom* axiom_;
-    const D* def_;
-};
+    /// @name name
```
fodinabor, line 111: ? :)

dialects/core/normalizers.cpp
```diff
@@ -18,17 +121,13 @@ constexpr std::array<std::array<u64, 2>, 2> make_truth_table(bit2 sub) {
 }
 
 // clang-format off
-
-// we rely on dependent lookup, so these cannot be overloads, but instead have to be
-// template specializations
+// We rely on dependent lookup, so these cannot be overloads, but instead have to be template specializations.
```
fodinabor, line 124: I believe this comment implies, that the `rop` and `rcmp` overloads have to be specializations as well.
Though, I am not quite sure, whether this is still needed, since we don't have this in the Thorin base anymore.

fodinabor (commented):
Generally looks great!
Didn't look at every line but still added some random comments here and there..

dialects/core/core.thorin
leissa: Good question. XD

Maybe rewrite rules can come to rescue. We have a similar situation in `%autodiff` where dialects unknown to `%autodiff` must be able to communicate to `%autodiff` that it has a really cool derivative for `%foo.bar` available:

```
.rule %autodiff.grad %foo.bar -> %foo.d_bar
```
And back to `%core.trait`:
```
.rule %core.trait %foo.Bar -> 42
```

NeuralCoder3: For now (and more generally), the dialect E that should be extended should probably make a map available and an extension function (similar to optimizations).
Then dialect A (or dialect C that depends on A and E) can register things there.

leissa:
I incorporated your comments. So unless this `op_malloc` thing is a show stopper, we could merge.

With my latest commit I added some doxygen and did a bit of refactoring to make the anatomy for axiom names hopefully a bit clearer. See `axiom.h`. If you have any objections, let me know.

leissa:
My latest commit resolves @fodinabor's remaining concern, thus I'm merging.

PR #88, opened 2022-09-28
Dependent direct plugin

NeuralCoder3:
The first step of AD merge:

A newer version of the ds2cps2ds dialect.
Now with dependent axioms.
Additionally, the beta reduction force axiom was moved from the
debug dialect to the (new) tool dialect.

Overall, the modularity and structure of the dialect were improved.
There are two main phases:
1. Convert direct style functions to cps function
  1.1. Generate cps functions with the same semantics
  1.2. Replace call sites of the ds function with wrapped calls to the cps functions
2. For ds calls of cps functions `cps2ds fun args`, restructure the program
    and move the calls to the cps position. Thus, eliminating the axiom call.

Note:
Curried ds functions and functions on type level are excluded. 

TODO:
- [x] resolve bug of disabled direct tests (newly introduced with the new PR)
- [x] cleanup (comments, includes)
- x resolve old var bug (delayed to next pull request -- #98)


NeuralCoder3:
Do we really need the tool dialect? At least `%tool.set_filter` should now be obsolete, no?
What about `%tool.force_type`? Isn't this a hack to fight a bug within Thorin itself? IMHO we should instead fix the bug within Thorin instead of committing a workaround.

_Originally posted by @leissa in https://github.com/AnyDSL/thorin2/pull/91#discussion_r985283950_
      


The `set_filter` is obsolete and I will remove it.
The forced beta reduction is currently used and it does not work automatically without it.
I could try if a filter set to true might be enough to fix it. Although this moves responsibility to the user who
needs to remember to set filters of type functions.
The easiest would be that thorin eagerly applies beta reduction on type-level functions.

leissa:
> The forced beta reduction is currently used and it does not work automatically without it. I could try if a filter set to true might be enough to fix it. Although this moves responsibility to the user who needs to remember to set filters of type functions. The easiest would be that thorin eagerly applies beta reduction on type-level functions.

Yes, this was my point. This seems to be more like a bug within Thorin itself. Fixing this should make the forced beta reduction superfluous, no?

NeuralCoder3:
It is not necessarily a bug. But rather an unintended design decision.
Yes, fixing the beta reduction (or equivalence check) will make it superfluous.

leissa:
> It is not necessarily a bug. But rather an unintended design decision. Yes, fixing the beta reduction (or equivalence check) will make it superfluous.

alright. Will look into that. The issue is already in `dialect.thorin`, if I use the alternative definition of this axiom, correct?

NeuralCoder3:
> alright. Will look into that. The issue is already in `dialect.thorin`, if I use the alternative definition of this axiom, correct?

Yes, if you swap line 30 and 31 (remove the `tool.force_type`), the tests will fail.

NeuralCoder3:
> Check out `beta_red_type_apps`.

The commit looks good and solves the issue.

leissa:
> > Check out `beta_red_type_apps`.
> 
> The commit looks good and solves the issue. Should the branch be merged into this branch or using a separate PR?

I justed merged.

leissa:
So, I just disabled this non-terminating Windows workflow. Should we merge or is someone with Windows around who can investigate what the heck is going wrong here? But to me it really looks like a bug in the Windows tool chain.

leissa:
Let's merge and open an issue for the Windows thing. Maybe @fodinabor can investiage when he's back from vacation on his Windows machine?

PR #90, opened 2022-09-29
Feature/rules

leissa:
This is what @michel-steuwer and I hacked together, if you are curious. You can now so sth like this:
```
.rule foo (x.Nat) = id x => x;
```
Since these rules are not actually used anywhere, I marked this as draft and just using this PR as an ongoing discussion.


NeuralCoder3:
Does any evaluation happen inside of the rules?

leissa:
> Does any evaluation happen inside of the rules?

just normal normalization.

PR #92, opened 2022-09-29
Feature/refly

leissa:
New reify/reflect dialect: `refly`
    
The `debug` dialect has been merged into this as these debugging features are also some kind of reflection.

The reflect/reify capabilities are rn more a demo but a good starting point, if we want to pick up on this later.  

The `%refly.gid` function is sth that we need for the rules #90.

Check out the `lit/refly/refine.thorin` demo for a glimpse of the insanity that is doable with this Oo


NeuralCoder3:
We probably want to keep a simple dialect similar to debug:
A small dialect that demonstrates how dialects are created, registered, ...

NeuralCoder3:
I think you used reify and reflect the right way around. But I am not sure.
Maybe that is an indication that it could be confusing. 
I would suggest naming the operators `quote` (forall T, T -> Code) and `unquote` (forall T, Code -> T).

dialects/refly/refly.thorin
```diff
@@ -0,0 +1,47 @@
...
+///
+/// Converts a `%%refly.Code` back to a Thorin expression.
+.ax %refly.reflect: Π T: * -> %refly.Code -> T, normalize_reflect;
+///
+/// ## Inspect
+///
+/// ### %refly.gid
+///
+/// Retuns the internal thorin::Def::gid of the argument.
+.ax %refly.gid: Π T: * -> T -> .Nat, normalize_gid;
```
NeuralCoder3, line 32: Isn't the gid very prone to changes? It's basically a new number after every optimization.
Can you elaborate on the use case?

NeuralCoder3:
A match would be good to inspect the generated `Code` object.
One way would be a comparison operator with a constructor.

dialects/refly/refly.thorin
```diff
@@ -0,0 +1,47 @@
...
+/// * `perm`: Permanently debugs at every construction.
+///   Gets removed at codegen preparation in thorin::refly::RemoveDbgPerm.
+.ax %refly.dbg(tmp, perm): Π T: * -> T -> T, normalize_dbg;
+/// 
+/// ## Manipulate
+/// 
+/// ### %refly.refine
+///
+/// Sets the `i`th operand of the reified Code `e` to `x`.
+.ax %refly.refine: Π T: * -> [e: %refly.Code, i: .Nat, x: T] -> %refly.Code, normalize_refine;
```
NeuralCoder3, line 47: An alternative approach (with larger meta programming capabilities in mind) would be a `getOps` and `rebuild (exp, ops)` axiom.
It might also be strange to insert real code into quoted terms instead of working the quoted world with quoted terms.

dialects/refly/normalizers.cpp
```diff
@@ -0,0 +1,44 @@
...
+const Def* normalize_reify(const Def*, const Def*, const Def* arg, const Def* dbg) { return do_reify(arg, dbg); }
+
+const Def* normalize_reflect(const Def*, const Def*, const Def* arg, const Def*) { return do_reflect(arg); }
+
+const Def* normalize_refine(const Def*, const Def* callee, const Def* arg, const Def* dbg) {
+    auto& world       = arg->world();
+    auto [code, i, x] = arg->projs<3>();
+    if (auto l = isa_lit(i)) {
+        auto def = do_reflect(code);
+        return do_reify(def->refine(*l, x), dbg);
```
NeuralCoder3, line 31: This only works if the quoted term is well typed. Do we want the invariant that quoted terms always need to be valid?

dialects/refly/normalizers.cpp
```diff
@@ -0,0 +1,44 @@
+#include "thorin/world.h"
+
+#include "dialects/refly/refly.h"
+
+namespace thorin::refly {
+
+static_assert(sizeof(const Def*) <= sizeof(u64), "pointer doesn't fit into Lit");
+
+static const Def* do_reify(const Def* def, const Def* dbg = {}) {
+    return def->world().lit(type_code(def->world()), (u64)def, dbg);
```
NeuralCoder3, line 10: How does this work? Or rather what does it do exactly?

leissa:
> We probably want to keep a simple dialect similar to debug: A small dialect that demonstrates how dialects are created, registered, ...

good point. maybe we should add a "demo" dialect or so with one operation.

leissa:
> I think you used reify and reflect the right way around. But I am not sure. Maybe that is an indication that it could be confusing. I would suggest naming the operators `quote` (forall T, T -> Code) and `unquote` (forall T, Code -> T).

Both reify and reflect are pretty well-established terms. E.g.:
* https://hackage.haskell.org/package/reflection-2.1.6/docs/Data-Reflection.html
* https://docs.scala-lang.org/overviews/reflection/symbols-trees-types.html

It's a bit philosophical whether quote/unquote from Scheme/LISP is basically the same thing - but I'd argue that reify/reflect is the more common terminology for this.

dialects/refly/refly.thorin
leissa: The use case is this:
```
.rule _ a + b = b + a if %refly.gid(b) < %refly.gid(a)
```
This is currently hard-coded in the normalizers of `%core` and is one of the things we may remove by the rule magic.

dialects/refly/refly.thorin
leissa: I just wanted to have one simple operation that allows us to manipulate code. If we are picking up on this - sure - we basically need to rebuild the important parts of the C++-API with `get_ops`, `rebuild`, `stub`, etc. But I think for now, the `%refly.refine` operation is a pretty simple example to manipulate sth.

leissa: Initially I thought, that it's cool to showcase how you can mix reified and actual expressions, but yes, that is too much. I changed the signature accordingly.

leissa:
So, unless you persistent on the demo dialect or have other objections, can we merge?

NeuralCoder3:
> So, unless you persistent on the demo dialect or have other objections, can we merge?

Yes, we can merge. The demo dialect can (and should) be re-added in another PR.

Issue #93, opened 2022-09-30
Add demo dialect


leissa:
Just that we don't forget about this one ...

NeuralCoder3:
Solved by #94

PR #94, opened 2022-09-30
Demo dialect

NeuralCoder3:
A simple demo dialect without any passes or fancy stuff.

The dialect just adds the axiom
`%demo.constIdx: Π [n:.Nat] -> .Idx n`
and a single test case.

The axiom is resolved to `42` by its normalizer.

This PR addresses #93


leissa:
cool :)
Maybe we should agree on a coding style for Thorin. I see slightly different styles used. But let's postpone this discussion for now.

Issue #95, opened 2022-10-04
Function used as a type but is in fact a term


NeuralCoder3:
Observed in `lit/direct/ds2cps_ax_cps2ds_dependent_short.thorin`
Command line: `./build/bin/thorin -d direct -d tool ./lit/direct/ds2cps_ax_cps2ds_dependent_short.thorin --output-thorin -`

Issue:
```
:4294967295: error: 'Π n_167425: .Nat → .infer (<nullptr>)' used as a type but is in fact a term 
```

High-level description:
We constructed a wrapper around `.Idx`

Suspected problem:
Idx has problems with the duality of fundamental integer and being applied as a function.

leissa:
Suffers also from #106. 
* changing the literal to `0:.Idx n` and
* adding a return type to `lam_idx` fixes the test case.

see also #103:

Work is underway to add type inference.

leissa:
Marked as enhancement as the only remaining real problem is the type inference that is not yet properly implemented.

leissa:
Main issues are resolved. Type inference has improved and will continue to improve in the future.

PR #96, opened 2022-10-05
Refactor/idx as fun

leissa:
`.Idx` is now a constant of type `.Nat -> *`.

Unfortunately, the prec-magic in the output is pretty broken. For this reason, I needed to port a lot of tests from `(.Idx 123)` to `.Idx 123`.

Rmark: We should really fix the prec magic before adding more and more test cases Oo


leissa:
Maybe this fixes #95

leissa:
I also finally refactored:
* `type_int_` -> `type_int`
* `lit_int_` -> `lit_int`
So, I hope you have ported your branches in the meantime :)

NeuralCoder3:
> I needed to port a lot of tests from `(.Idx 123)` to `.Idx 123`.

This seems odd. Additional parentheses should not make a program invalid.
Especially since these parentheses are usually needed in other contexts in thorin (as opposed to other functional languages (Haskell, coq, OCaml, ...) where they are commonly omitted).

leissa:
> > I needed to port a lot of tests from `(.Idx 123)` to `.Idx 123`.
> 
> This seems odd. Additional parentheses should not make a program invalid. Especially since these parentheses are usually needed in other contexts in thorin (as opposed to other functional languages (Haskell, coq, OCaml, ...) where they are commonly omitted).

No, I needed to port the `CHECK-DAG` part not the actual test case.

NeuralCoder3 (approved):
Looks good.

PR #100, opened 2022-10-10
This restricts the grammar a little bit.

leissa:
Production
```
e -> d e
```
was removed. Instead we now have:
```
de -> d* e
```
And some very specific places where `de` is allowed: 
```
e -> { de }
o -> = de ;
```
This has the effect that a typical error like this: 
```
.let x = foo 1 2 3 /* missing semicolon */
.let y = bar 4 5;
```
is now a syntax error.


NeuralCoder3:
This would affect something like
``` 
.ax %matrix.transpose: Π [kl: «2: .Nat; .Nat», T: *] -> 
    .let (k,l) = kl;
    %matrix.Mat (2,(k,l),T) -> %matrix.Mat (2,(l,k),T), normalize_transpose;
``` 
right?

leissa:
> This would affect something like
> 
> ```
> .ax %matrix.transpose: Π [kl: «2: .Nat; .Nat», T: *] -> 
>     .let (k,l) = kl;
>     %matrix.Mat (2,(k,l),T) -> %matrix.Mat (2,(l,k),T), normalize_transpose;
> ```
> 
> right? Because now `.let` is only allowed at the beginning but not in between expressions.

This is correct. But what you can do instead is the following:
```
.ax %matrix.transpose: Π [kl: «2: .Nat; .Nat», T: *] -> 
    {
        .let (k,l) = kl;
        %matrix.Mat (2,(k,l),T) -> %matrix.Mat (2,(l,k),T)
    }, normalize_transpose;
```

leissa:
yes, but I do see the value of putting `.let`s in between stuff without opening a new block. Yet, missing semicolons are really annoying errors. Any other ideas?

NeuralCoder3:
> yes, but I do see the value of putting `.let`s in between stuff without opening a new block. Yet, missing semicolons are really annoying errors. Any other ideas?

Maybe introduce an additional inline-let `e -> let p = e in e` .
The let with a semicolon in an expression looks weird in some circumstances (like the one above).

leissa:
What about a production
```
e -> .with de 
```
Then, you could do this:
```
.ax %matrix.transpose: Π [kl: «2: .Nat; .Nat», T: *] -> 
    .with 
        .let (k,l) = kl;
    %matrix.Mat (2,(k,l),T) -> %matrix.Mat (2,(l,k),T), normalize_transpose;
```

NeuralCoder3:
> What about a production
> 
> ```
> e -> .with d* .in e 
> ```

That is basically the same syntax as with a block. Which is also fine.
Also, I think is not that common. Therefore, we probably do not need a special notation. (In the case above, a deeper intro-pattern avoids the need to destruct kl)

NeuralCoder3 (approved):

leissa:
> > What about a production
> > ```
> > e -> .with d* .in e 
> > ```
> 
> That is basically the same syntax as with a block. Which is also fine. Also, I think is not that common. Therefore, we probably do not need a special notation. (In the case above, a deeper intro-pattern avoids the need to destruct kl)

yes, but in contrast to `{ de }` the `.with`/`.in` expression doesn't open a new scope. So:
```
{
    .let x = ...;
    x;
}
...
use(x); // error
```
in contrast to
```
.with
    .let x = ...;
.in
    x;
...
use(x); // fine
```

leissa:
Let me just merge this like it is now. If we need sth like the`.with`-expr we can still add this later on.

PR #104, opened 2022-10-12
feature/nat_ops

leissa:
Just sth we needed when working on #90 and is potentially useful for everyone to have.

### DIFF
--- a/dialects/core/core.thorin
+++ b/dialects/core/core.thorin
@@ -12,6 +12,16 @@
 ///
 .ax %core.Real: .Nat -> *;
 ///
+/// ## Nat Operations
+///
+/// ### %core.ncmp
+///
+.ax %core.ncmp(e, ne, l, le, g, ge): Π [.Nat, .Nat] -> .Bool, normalize_ncmp;
+///
+/// ### %core.nop
+///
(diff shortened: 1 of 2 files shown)


NeuralCoder3:
Why are nop as axioms needed?
Isn't the semantics 
```
%core.nop.add (a,b)
```
the same as 
```
.let a_i32 = %core.bitcast (I32, .Nat) a;
.let b_i32 = %core.bitcast (I32, .Nat) b;
.let c_i32 = %core.wrap.add (0,_32) (a_i32,b_i32);
%core.bitcast (.Nat, I32) c_i32
```
?
Even the normalization should be (nearly) the same.

leissa:
It just makes our lives easier to directly have `Nat`-based operations. I think it makes sense to extend the special axiom namespace to work for all kind of entities - as I've already suggested. Then, we could define those things like this:
```
.lam %core.nop.add (a: .Nat, b: .Nat) -> .Nat =
    .let a_i32 = %core.bitcast (I32, .Nat) a;
    .let b_i32 = %core.bitcast (I32, .Nat) b;
    .let c_i32 = %core.wrap.add (0,_32) (a_i32,b_i32);
    %core.bitcast (.Nat, I32) 
```

Issue #101, opened 2022-10-10
Windows linker going bonkers


leissa:
After a recent merge the Windows linker doesn't terminate anymore in the `Debug` build.

leissa:
fixed with #107

Issue #106, opened 2022-10-12
k:.Idx n doesn't make sense


leissa:
My latest commit fixed some UB which materialized in the CI that this test here
```
ds2cps_ax_cps2ds_dependent2.thorin
```
failed to work *sometimes* on Windows.

Problem is: A literal `42:.Idx n` doesn't make sense and should have been a type error (which it is now). It caused UB in some normalizers (which I have fixed). 

One edge case is: `0:.Idx n`. This kind of makes sense, as `0:.Idx 0` actually means `0:.Idx 2^64`. So this literal is guaranteed to exist - but this is still kind of odd Oo Should we really allow that?

For now I disabled the test case above, as it makes use of this erroneous index.

Maybe related to #101.
Also, check out #105.

NeuralCoder3:
A cast s2s/u2u should also fix the problem is probably the correct fix.

I think we should allow `0:.Idx n` as it might be helpful to have a zero sometimes and it exists from a theoretical point of view.

leissa:
@NeuralCoder3: can you fix the `ds2cps_ax_cps2ds_dependent2.thorin` test case? Then, I can close this one here already.

Issue #105, opened 2022-10-12
valgrind for lit tests


leissa:
We should definitely enable valgrind for lit tests. It's just too easy with C++ to make some stupid error. Currently, I'm investigating some bugs that wouldn't have slipped through with valgrind enabled.

fodinabor:
In theory does `lit` provide a `--vg` parameter that can be used to say, please run the tests inside with valgrind.
That also works on my machine (probably won't on Windows..?) but I'm not sure how often we want to use that as the performance is obviously awful... it also runs clang or whatever else we specify to run in the `RUN` lines with valgrind (although thorin clearly dominated the testing time..).
Timings of running the lit tests on my 16 thread Ryzen 7 5850u:
| Config | No --vg | --vg |
| Debug | 20min | 22s |
| Release | 200s | 17s |

So I'd provide either a CMake option to add the parameter or just provide a `check-valgrind` target or so?
Not sure if we want to run that in the CI though... maybe in the Release build?
What are your thoughts?

fodinabor:
given that we're down to 200s total, when building Thorin in Release, I would claim valgriding clang does not hurt *too* much.

I assume we could still build our own solution to just valgrind thorin: since we set the value of `%thorin` in the lit config ourselves, we could also have some config option that replaces that with `valgrind ... /path/to/thorin` instead of just `/path/to/thorin`.
That obviously won't change much for that in debug thorin is the dominating factor.

leissa:
> since we set the value of `%thorin` in the lit config ourselves, we could also have some config option that replaces that with `valgrind ... /path/to/thorin` instead of just `/path/to/thorin`. 

I really like this approach.

> That obviously won't change much for that in debug thorin is the dominating factor.

I don't really get why this takes so long. `make check` takes 21s on my old workstation here. Now using valgrind on Thorin alone would probably have an impact of an order of magnitude or so but not 2 orders of magnitudes???

fodinabor:
Release with `%thorin` set to `valgrind /path/to/thorin` takes 75s.
Debug 16m. (haven't rerun and averaged these numbers obvs..)

leissa:
So my vote would be to enable valgrind for release builds in GitHub CI and by enabling it simply using `valgraind /path/to/thorin`.

My guess is that this dynamic loading of plugins is really, really slow with valgrind.

PR #107, opened 2022-10-12
reenable Win Debug

leissa:


leissa:
Indeed fixes #101.
See also #106, #105.

Issue #10, opened 2022-02-22
Erroneous bitcast for indefinite array allocation


NeuralCoder3:
Allocations with a statically unknown size `T::nat` lead to `<TODO>` in LLVM code.

For instance, when an array of the same size as an indefinite array `«⊤∷nat; r32»` is created in Thorin,
invalid LLVM code is produced by the code generation.

Using Alloc2Malloc:
Before optimizations:
```
_1176: [:mem, :ptr («3∷nat; r32», 0∷nat)] = :alloc («3∷nat; r32», 0∷nat) mem_1175;
_1190: :ptr («⊤∷nat; r32», 0∷nat) = :bitcast (:ptr («⊤∷nat; r32», 0∷nat), :ptr («3∷nat; r32», 0∷nat)) _1176#1∷i1;
...
_1159: [:mem, :ptr («⊤∷nat; r32», 0∷nat)] = :alloc («⊤∷nat; r32», 0∷nat) _1158;
```
After optimizations
```
  _2412: [:mem, :ptr («3∷nat; r32», 0∷nat)] = :malloc («3∷nat; r32», 0∷nat) (mem_2410, 12∷nat);
  _2445: i64 = :bitcast (i64, nat) ⊤∷nat;
  _2449: i64 = :Wrap_mul (3∷nat, 0∷nat) (4∷i64, _2445);
  _2450: nat = :bitcast (nat, i64) _2449;
  _2452: [:mem, :ptr («⊤∷nat; r32», 0∷nat)] = :malloc («⊤∷nat; r32», 0∷nat) (_2412#0∷i1, _2450);
```
and the corresponding LLVM code:
```
  %_2755.i8 = call i8* @malloc(i64 12)
  %_2755 = bitcast i8* %_2755.i8 to [3 x float]*
  %_2788 = bitcast i64 <TODO> to i64
  %_2792 = mul nuw nsw i64 4, %_2788
  %_2793 = bitcast i64 %_2792 to i64
  %_2795.i8 = call i8* @malloc(i64 %_2793)
```

Without Alloc2Malloc:
After optimizations:
``` 
_2375: [:mem, :ptr («3∷nat; r32», 0∷nat)] = :alloc («3∷nat; r32», 0∷nat) mem_2374;
_2377: [:mem, :ptr («⊤∷nat; r32», 0∷nat)] = :alloc («⊤∷nat; r32», 0∷nat) _2375#0∷i1;
```
LLVM: (no malloc or alloc is generated)
[...]

PR #110, opened 2022-10-13
Arr opt

### LINKED ISSUE #109: Performance depends on the size of Arr/pack
Here Thorin is doing things `100000000` times.
```
.import mem;

.let i32 = .Idx 4294967296;
.let size = 100000000:.Nat;

.cn .extern main [mem: %mem.M, argc: i32, argv : %mem.Ptr («⊤:.Nat; %mem.Ptr («⊤:.Nat; .Idx 256», 0:.Nat)», 0:.Nat), return : .Cn [%mem.M, i32]] = {
    .let (alloc_mem, arr) = %mem.alloc (<<size; i32>>, 0) (mem);
    .let lea = %mem.lea (size, <size; i32>, 0) (arr, 0:(.Idx 100));
    .let (load_mem, val) = %mem.load (i32, 0) (alloc_mem, lea);

    return ( load_mem, 1:i32 )
};
```

leissa:
Trying to fix #109.


NeuralCoder3:
Does this branch solve the problem and is ready?

leissa:
No, it's only half of the equation. One problem was whether checking for alpha-equiv happens only optimistic or not. I.e. `t1 == t2` **must** yield `true`, otherwise sth is broken. But there is also this `is_uniform` thing, where alpha-equiv is only optionally but we rely on that.

Anyway, I'm currently working (again) on the type inference which is strongly geared into the alpha-equiv and I'm looking again at this issue from this angle.

leissa:
After our recent discussion in Discord, I looked once again into the changes here. One change was the problematic rule:
```
(a, b) -> ‹2; a›    if alpha-equiv(a, b).
```
Which is replaced by this PR with:
```
(a, b)#i (x, y, z)
* a: A = .Cn [X, Y, Z]
* b: B = .Cn [X, Y, Z]
```
While this does not yet fully fixes #10, it makes things better.

PR #111, opened 2022-10-13
Closure conversion

NeuralCoder3:
The closure conversion dialect.
Allows for the handling of higher-order functions including interconnectivity with memory operations.


leissa:
What is this `.DS_Store`? There are two binary files in the commit Oo

NeuralCoder3:
> What is this `.DS_Store`? There are two binary files in the commit Oo

`.DS_Store` are temporary files that macOS adds to every folder.
They were probably accidentally added in `e9d2fe` and can be removed.

PR #113, opened 2022-10-13
make things a bit more consistent with the other dialects

leissa:
* autodiff.autodiff_type -> autodiff.AD
* autodiff.autodiff -> autodiff.ad
* autodiff.tangent_type -> autodiff.Tangent
* --output-thorin -> -o
* remove some superfluous comments

### DIFF
--- a/dialects/autodiff/autodiff.h
+++ b/dialects/autodiff/autodiff.h
@@ -10,7 +10,7 @@ inline const Def* op_autodiff(const Def* fun) {
     World& world = fun->world();
     // We rely on the normalized thorin convention that all arguments in functions are grouped.
     // `cn[[args], cont:=cn[returns]]`
-    return world.app(world.app(world.ax<autodiff>(), fun->type()), fun);
+    return world.app(world.app(world.ax<ad>(), fun->type()), fun);
 }
 
 inline const Def* op_zero(const Def* A) {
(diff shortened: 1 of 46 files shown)


leissa:
There is also the convention that the C++ c'tor that constructs a type uses `type` as **pre**fix such as `type_idx` or `type_nat`. There are a couple of C++ type c'tors within `autodiff` that don't follow this convention but I didn't want to change too much.

NeuralCoder3:
> There is also the convention that the C++ c'tor that constructs a type uses `type` as **pre**fix such as `type_idx` or `type_nat`. There are a couple of C++ type c'tors within `autodiff` that don't follow this convention but I didn't want to change too much.

I am fine with more renaming.
But the `autodiff_type_fun` (fun was added to avoid issues with the axiom name) is a type transformer for autodiff as opposed to creating an autodiff type.
Therefore, it is conceptually different from `type_int`.

NeuralCoder3 (approved):

leissa:
> > There is also the convention that the C++ c'tor that constructs a type uses `type` as **pre**fix such as `type_idx` or `type_nat`. There are a couple of C++ type c'tors within `autodiff` that don't follow this convention but I didn't want to change too much.
> 
> I am fine with more renaming. But the `autodiff_type_fun` (fun was added to avoid issues with the axiom name) is a type transformer for autodiff as opposed to creating an autodiff type. Therefore, it is conceptually different from `type_int`.

That was also my suspicion. so let's keep it like this for now :)

Issue #118, opened 2022-10-14
Closure Conversion - LLVM Emission Error


NeuralCoder3:
In some cases, the emitted LLVM Code writes and loads functions from a wrong slot resulting in an infinite loop.

Example:
https://github.com/NeuralCoder3/thorin2/blob/closure-test/lit/clos/bind.thorin

Commands:
```bash
%thorin -d affine %s --output-ll %t.ll -o -
clang++ lit/clos/lib.cpp %t.ll -o %t -lc -Wno-override-module
``` 

Observed bahvior:
An infinite stream of `99` is printed.

Expected behavior:
A countdown from `99` to `0`.

PR #124, opened 2022-10-24
Bugfix/bind

leissa:
fix for #118


leissa:
Note that placing allocas not within the entry block may cause severe performance regressions within LLVM.

PR #131, opened 2022-10-28
Feature/math

leissa:
# Rework of floating point support in a new dialect: `math`

Impala branch [here](https://github.com/AnyDSL/impala/tree/feature/math).

## TODO

* [x] math.thorin
* [x] finish first version of normalizers
* [x] port `%core` real stuff to new `%math` API 
* [x] remove old `%core` real stuff
* [x] implement llvm codegen
* [x] mode for `%math.cmp`
* [x] rework `*.conv` normalizations

## Change Log

* `%core.Real w` -> `%math.F (p, e)`
   Originally floating point types were named `Real` to not collide with the C++ keyword.
   But now we have a whole dialect about floating point so I think it's pretty clear what the type constructor `%math.F` is doing. On top of that, the type constructor now expects number of precision and exponent bits to allow for `bfloat16` and other crazy formats. For now, however, constant folding and the backend only works for `F32`, `F64` and partially for `F16`.

* `%core.rop` -> `%math.arith`
* `%core.rcmp` -> `%math.cmp`
* `core::WMode` -> `core::Mode`
* `core::RMode` -> `math::Mode`
* `%core.conv` has been split into:
    * `%core.conv.u2u`
    * `%core.conv.s2s`
    * `%math.conv.f2f`
    * `%math.conv.s2f`
    * `%math.conv.u2f`
    * `%math.conv.f2s`
    * `%math.conv.f2u`
* In order to allow for easy type inference in the future the non-inferable `mode` argument has been curried just in front of the actual argument tuple - leaving the inferable argument tuple as the first curried argument.
  E.g.:
  ```
  .ax %math.arith(add, sub, mul, div, rem):
      Π [p: .Nat, e: .Nat] -> .Nat -> [%math.F (p, e), %math.F (p, e)] -> %math.F (p, e), normalize_arith;
  ```
  For this reason the signature of the `*.conv`s has changed from
  ```
  Π [dw: .Nat, sw: .Nat] -> %core.Real sw -> %core.Real dw
  ```
  to
  ```
  Π [sp: .Nat, se: .Nat] -> Π [dp: .Nat, de: .Nat] -> %math.F (sp, se) -> %math.F (dp, de)
  ```
  Note that the order of source and destination has switched and is now curried.
* All other signatures within `%core` remain as they are for now.
* New axioms:
    * `%math.x` (min/max)
    * `%math.tri`
    * `%math.pow`
    * `%math.rt` (square and cubic root)
    * `%math.exp`
    * `%math.er`
    * `%math.gamma`
* Codegen for `%core.nop` and `%core.ncmp`. (`%core.trait` now uses `%core.nop`).
* Normalizations got a face lift.
* ~~Normalizations such as constant folding have not been implemented for most new `%math` operations.~~ Constant folding has been ipmlemented; other normalizations such as `pow(x, 0) -> 1` are missing rn. But this is something we can add bit by bit as needed.
* Rn, the new math operations directly generate LLVM code. I think it would be better to remove this - at least for most operations - and instead lower them to usual functions calls. This is what's happening anyway. However, to pull that off nicely, we need a general mechanism to lower axioms into sth else late in the optimizations pipeline simply by "linking" the proper impl of an axiom into the module or so.
* I haven't tested all operations. We can add that later on.


lit/math/exp.thorin
```diff
@@ -0,0 +1,24 @@
...
+.import math;
+.import mem;
+.import core;
+
+.let _32 = 4294967296;
+.let I32 = .Idx _32;
+.let f32 = (23,  8);
+.let f64 = (52, 11);
+.let F32 = %math.F f32;
+.let F64 = %math.F f64;
```
NeuralCoder3, line 15: This code (at least the f F definitions) is so common that we might want it directly in the dialect thorin files.

dialects/math/math.thorin
```diff
@@ -0,0 +1,191 @@
...
+/// ### %math.rt
+///
+/// [Square](https://en.wikipedia.org/wiki/Square_root) and [cube root](https://en.wikipedia.org/wiki/Cube_root):
+///
+/// | Name           | Meaning     | Semantics          |
+/// |----------------|-------------|--------------------|
+/// | `%%math.rt.sq` | square root | \f$\sqrt{x}\f$     |
+/// | `%%math.rt.cb` | cube root   |\f$\sqrt[3]{x}\f$   |
+///
+.ax %math.rt(sq,cb): Π [p: .Nat, e: .Nat] -> .Nat -> [%math.F (p, e), %math.F (p, e)] -> %math.F (p, e), normalize_rt;
```
NeuralCoder3, line 110: Later on (when we support definitions with namespaces in dialects), we should just define sqrt using pow to avoid duplication.

lit/math/exp.thorin
leissa: I'd like to have this later on in the `%math` namespace. Once we allow this for everything and not just axioms:
```
.let %math.f32 = (23, 8);
...
```

PR #132, opened 2022-10-29
_and, _xor, _or -> and_, xor_, or_

leissa:
I think some markdown stuff is having problems with an underscore prefix.


leissa:
This indeed fixes the failing Doxygen tests.

Issue #126, opened 2022-10-25
Closure Conversion - Should work for functions without %mem.M type


leissa:
This supersedes #119.

PR #138, opened 2022-11-06
Bugfix/clos conv

### LINKED ISSUE #117: Closure Conversion - higher order return
``` 
1E) '([], ((.Idx 4294967296, aug_pow_else_745690_660849_352178, _660885_351873):closure_type_352061, ([], aug_pow_then_745900_661112_352340, ()):closure_type_352416)#_352541#2:(.Idx 3))' of type 
2E) '      (.Idx 4294967296, aug_pow_else_745690_660849_352178, _660885_351873):closure_type_352061, ([], aug_pow_then_745900_661112_352340, ()):closure_type_352416)#_352541#1:(.Idx 3)' of domain 

1T) '[★,                   ((.Idx 4294967296, aug_pow_else_745690_660849_352178, _660885_351873):closure_type_352061, ([], aug_pow_then_745900_661112_352340, ()):closure_type_352416)#_352541#0:(.Idx 3)]' to 
2T) '[closure_type_352015, ((.Idx 4294967296, aug_pow_else_745690_660849_352178, _660885_351873):closure_type_352061,  ([], aug_pow_then_745900_661112_352340, ()):closure_type_352416)#_352541#0:(.Idx 3)]'
      ^ only here
``` 

This behavior is exhibited for instance in `pow_autodiff_eval_filter2_eval.thorin`.

The program is a differentiated version of pow and as such a minimal recursive function that folds continuations inside each other in the return continuation.

leissa:
Just a draft. Trying to fix #117 and #126 and along the way port the phases from `thorin::Phase` to `thorin::RWPhase` to reduce some boilerplate.

### DIFF
--- a/dialects/clos/clos.cpp
+++ b/dialects/clos/clos.cpp
@@ -131,7 +131,6 @@ const Sigma* isa_clos_type(const Def* def) {
 }
 
 Sigma* clos_type(const Pi* pi) { return pi2clos(pi); }
-    //return ctype(pi->world(), pi->doms(), nullptr)->as_nom<Sigma>(); }
 
 const Pi* clos_type_to_pi(const Def* ct, const Def* new_env_type) {
     assert(isa_clos_type(ct));
--- a/dialects/clos/clos.cpp
+++ b/dialects/clos/clos.cpp
@@ -20,36 +20,22 @@
 
 namespace thorin::clos {
 
-template<bool type>
-size_t env_idx(const Def* def) {
-    size_t n = def->num_projs();
-    if (n == 0) return 0;
-    auto first = def->proj(n, 0_s);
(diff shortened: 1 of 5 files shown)


NeuralCoder3:
I do not know the inner workings of the closure conversion pass.

But in general, it might be useful to solve this issue by adding a separate pass in mem or clos that propagates mem and adds it everywhere.

Together with a (fixed) version of reshape/scalerize that makes everything flat
and a (maybe separate from the previous) pass that reorders arguments,
this should be enough to fix this issue.

The added benefit would be that operations can request pseudo-memory nodes out of thin air. This allows optimizing more freely. 
The forced (partial) ordering could still be enforced with the pseudo-memory elements.
The memory-propagation pass (or an additional pass) would replace them.

Nevertheless, it might still be helpful to make closure conversion more resistant depending on its current state in a next step.

PR #139, opened 2022-11-07
Idx size

leissa:
Use the smallest available LLVM int type for Thorin .Idx types.

### DIFF
--- a/dialects/core/be/ll/ll.cpp
+++ b/dialects/core/be/ll/ll.cpp
@@ -128,12 +128,10 @@ std::string Emitter::id(const Def* def, bool force_bb /*= false*/) const {
 }
 
 static nat_t size2llbits(const Def* size) {
-    if (auto s = isa_lit(size)) {
-        if (*s == 0) return 64;
-        if (*s <= 0x0000'0000'0000'0002_u64) return 1;
-        if (*s <= 0x0000'0000'0000'0100_u64) return 8;
-        if (*s <= 0x0000'0000'0001'0000_u64) return 16;
-        if (*s <= 0x0000'0001'0000'0000_u64) return 32;
+    if (auto s = Idx::size2bitwidth(size)) {
+        nat_t r = std::bit_ceil(*s);
+        if (1 < r && r <= 8) return 8;
+        return r;
     }
     return 64;
 }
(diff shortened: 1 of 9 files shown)


fodinabor (commented):
Is `size2bitwidth` still in use or should it be cleaned up?

dialects/core/normalizers.cpp
```diff
@@ -576,11 +576,11 @@ const Def* normalize_trait(const Def*, const Def* callee, const Def* type, const
...
-        if (auto w = isa_lit(size)) {
-            if (*w == 0) return world.lit_nat(8);
-            if (*w <= 0x0000'0000'0000'0100_u64) return world.lit_nat(1);
-            if (*w <= 0x0000'0000'0001'0000_u64) return world.lit_nat(2);
-            if (*w <= 0x0000'0001'0000'0000_u64) return world.lit_nat(4);
+        if (auto s = isa_lit(size)) {
+            if (*s == 0) return world.lit_nat(8); // i64
+            if (*s <= 0x0000'0000'0000'0100_u64) return world.lit_nat(1);
+            if (*s <= 0x0000'0000'0001'0000_u64) return world.lit_nat(2);
+            if (*s <= 0x0000'0001'0000'0000_u64) return world.lit_nat(4);
```
NeuralCoder3, line 583: code duplication with  size2llbits -> just use the function and divide by 8.
Or more generally, refactor the functionality out.

dialects/core/be/ll/ll.cpp
```diff
@@ -498,24 +493,14 @@ std::string Emitter::emit_bb(BB& bb, const Def* def) {
...
-                    case  8: return std::to_string(lit->get< u8>());
-                    case 16: return std::to_string(lit->get<u16>());
-                    case 32: return std::to_string(lit->get<u32>());
-                    case 64: return std::to_string(lit->get<u64>());
-                    // clang-format on
-                    default: return std::to_string(lit->get<u64>());
-                }
-            } else {
-                return std::to_string(lit->get<u64>());
+            if (auto s = isa_lit(size)) {
```
NeuralCoder3, line 496: Code duplication with other points where s is converted to bits.

leissa:
> Is `size2bitwidth` still in use or should it be cleaned up?

This is still in use. `size2bitwidth` will give you 8 from 256 but `std::nullopt` from 255 whereas `size2llwidth` will give you 8 in both cases.

dialects/core/normalizers.cpp
leissa: yes, I stumbled over this myself. Problem is that different places are doing slightly different things. Here we are calculating the size in bytes. In the llvm backend we are mapping to llvm types - which is slightly different due to the existence of `i1`

dialects/core/be/ll/ll.cpp
leissa: This is hard to factor out due to the template arguments involved.

dialects/core/normalizers.cpp
leissa: I could have a methods which does this and in the llvm backend a function that wraps this and puts an extra case for `i1` in front of it.

leissa:
I'll try to refactor some things as discussed above.

dialects/core/be/ll/ll.cpp
NeuralCoder3: Alternatively, one could use a `typename std::conditional` construct for `uint8_t -> type` and use this.
I think this would be better as it isolated the different logic and moves the conversion with (semi-)magic numbers to a single point.

dialects/core/normalizers.cpp
NeuralCoder3: I think the common logic should be unified or at least grouped together.
Otherwise, we have very similar (at first glance not all differences are apparent) logic all over the place that depend on each other semantically.

leissa:
Okay, I think this should address your points.

NeuralCoder3 (approved):

PR #140, opened 2022-11-09
RWPhase that writes into the same world

leissa:
The old `RWPhase` needed to always rewrite everything as it filled a new world. This makes it difficult to **not** rewrite anything. With this change, you can simply opt to return the original def to stop recursing.

### DIFF
--- a/thorin/phase/phase.cpp
+++ b/thorin/phase/phase.cpp
@@ -9,10 +9,8 @@ void Phase::run() {
 }
 
 void RWPhase::start() {
-    for (const auto& [_, ax] : old_world().axioms()) rewrite(ax);
-    for (const auto& [_, nom] : old_world().externals()) rewrite(nom)->as_nom()->make_external();
-
-    swap(Phase::world_, new_world_);
+    for (const auto& [_, ax] : world().axioms()) rewrite(ax);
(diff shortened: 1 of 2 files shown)


NeuralCoder3 (approved):
Looks good. This should also reduce the "Old var still around" errors and make them easier debugable.

NeuralCoder3:
This does not solve the type axiom issue, right?
In principle, the order is still:
* operands of an app are rewritten, 
* the app is rebuild
* the app is rewritten

leissa:
> This does not solve the type axiom issue, right? 

what is the "type axiom issue"? XD

> In principle, the order is still:
     * operands of an app are rewritten,
     * the app is rebuild
     * the app is rewritten

yes, but at any point you can opt to do sth else, e.g.:
```c++
const Def* MyRWPhase::rewrite_structural(const Def* def) {
    if (auto app = def->isa<App>()) {
        if (is_weird(app)) return app; // stop recursing here
        if (is_crazy(app)) {
            auto callee = rewrite(app->callee()); // recursively rewrite callee
            auto arg = app->arg();                // don't recurse here for whatever reason
            return my_other_crazy_app(callee, arg);            
        }
        // note the fallthrough here
    }
    
    return Rewriter::rewrite_structural(def); // continue recursive rewriting with everything else
}
```

PR #142, opened 2022-11-10
Feature/lam expr

leissa:
Implements lambda and continuation *expressions*.

New/Updated syntax:

| Declaration | Expression |
|-|-|
| `.con` sym pat `=` e `;` | `.cn` pat `=` e |
| `.lam` sym pat `=` e `;` | `.lm` pat `=` e \|   `λ` pat `=` e |

This is a bit simplifed. Check out `langref.md` for the exact grammar.

I also started working on `.fn`/`.fun`, but I will do this in another PR.


NeuralCoder3 (commented):
I am not entirely convinced about `.con` as a name. But I do not have a better alternative right now.
Especially since the function in itself only becomes a continuation when applied.

Other than that, it looks good.

PR #146, opened 2022-11-19
bugfix: detetect recursive function types in axioms

leissa:
Bug fix for axioms regarding recursive function types.

### DIFF
--- a/lit/rec_pi.thorin
+++ b/lit/rec_pi.thorin
@@ -1,6 +1,5 @@
 // RUN: %thorin %s -o -
 
-.ax %phase.Pipeline: *;
 .ax %phase.Phase: *;
 
 .ax %phase.phase1: .Bool         -> %phase.Phase;
--- a/lit/rec_pi.thorin
+++ b/lit/rec_pi.thorin
@@ -1,0 +1,18 @@
+// RUN: %thorin %s -o -
+
+.ax %phase.Pipeline: *;
+.ax %phase.Phase: *;
+
+.ax %phase.phase1: .Bool         -> %phase.Phase;
+.ax %phase.phase2: []            -> %phase.Phase;
+.ax %phase.phase3: [.Nat, .Bool] -> %phase.Phase;
+
+.Pi PiPeline: *, %phase.Phase = PiPeline; // got the pun? XD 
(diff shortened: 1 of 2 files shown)


NeuralCoder3:
`.Pi PiPeline: *, %phase.Phase = PiPeline;`
is a notation for
`PiPeline := μ α. Phase → α`
which is handled as equirecursive type?

Therefore, a type that is automatically unfolded to a function that takes a phase — effectively resulting in functions taking arbitrarily many phases (a list of phases) and can only be handled using axioms.

leissa:
yes, 
```
Pi PiPeline: *, %phase.Phase = PiPeline;
```
is
```
PiPeline := μ α. Phase → α
```

I'd argue it's iso-recursive. The nominal tag in the type is so to say the tag where to place `unfold`s. If you manually unfold the type, Thorin will not handle them as equivalent.

Originally, I intended to have all nominal entities to be - well - nominally typed. Unfortunately, this didn't turn out to be enough. For this reason we are also checking for alpha-equiv even for nominals - which makes the term *nominal* debatable. 

Anyway, this is not restricted to axioms. Check out this example:
```
.Pi Hungry: *, .Nat = Hungry;
.lam hungry (n: .Nat) -> Hungry = hungry;

.let x = hungry 0;
.let y = hungry 0 1;
.let z = hungry 0 1 2;
.let w = hungry 0 1 2 3;
```
(From TaPL, 270).

NeuralCoder3:
> I'd argue it's iso-recursive. The nominal tag in the type is so to say the tag where to place unfolds. If you manually unfold the type, Thorin will not handle them as equivalent.

What is the correct way in our implementation to unify the types / let thorin recognize the convertibility (iota-reduction)?
E.g.
```
.ax %compile.Phase: *; 
.Pi PassPhase: %compile.Phase, %compile.Pass = PassPhase;
// this should be of type %compile.Phase but reports typing errors
.let optimization_phase = {
    %compile.pass_phase
        %core.partial_eval_pass
        %core.beta_red_pass
}
``` 
One trick would be to define `Phase` as `PassPhase` but this would introduce unwanted side effects.

leissa:
So, problem solved?

PR #147, opened 2022-11-21
Add new AddMem phase

fodinabor:
This adds a `%mem.M` to all nominals, continuations, and their calls.

The basic algorithm is as follows:
- use `Scheduler` to provide the information in which lam a node is scheduled to
- traverse in a depth-first (pre-order!) manner through the graph
- whenever a new mem is introduced (rewritten), set the lam's current mem to the new mem
- always use the current mem from the lam a node is scheduled to instead of the input mem
- remember always to visit preexisting input mems first either way, so that the order of mem's set as current mem is respected

I haven't updated the tests yet, since I'm unsure if we want to enable this phase in all cases?
Also, I assume it will be needed at a different spot, not in codegen prepare, but before closure conversion?


NeuralCoder3:
> I haven't updated the tests yet, since I'm unsure if we want to enable this phase in all cases?
> Also, I assume it will be needed at a different spot, not in codegen prepare, but before closure conversion?

Yes, I think only closure conversion should call this phase as part of its preparation. ([Example](https://github.com/NeuralCoder3/thorin2/blob/ho_codegen/dialects/clos/clos.cpp#L151))
And maybe also new dialects with passes that make use of the new functionality.
Therefore, it should not be part of the default mem pipeline.

leissa:
I just looked briefly into the impl. So, I have the impression, that you want to consider one top-level scope at a time. You can use `ScopePhase` for this.
 
I intended the `RWPhase` for any kind of recursive rewriting - what you are also doing here, but you are duplicating the logic. `RWPhase` inherits from `Rewriter` and already provides a `Rewriter::map` method and checks whether sth is already inside. Sth like this:

```c++
class AddMem : public RWPhase {
    const* Def rewrite_structural(const Def* def) override {
        if (auto app = def->isa<App>()) {
            if (auto lam = do_stuff(app)) {
                auto arg = rewrite(app->arg()); // recurse
                auto new_lam = /*...*/; // make our own stuff here; s.a. Rewriter::rewrite_nom
                return world().app(new_lam, arg);
            }
        }
        
        return Rewriter::rewrite_structural(def); // continue recursing for everything else
    }
}
```

If we need a `ScopeRWPhase` it would probably a good idea to also provide this within `phase.h`. Either way, I'd really like to have a common code base for this typical recursive rewriting stuff.

thorin/pass/pass.cpp
```diff
@@ -46,14 +46,14 @@ void PassMan::run() {
     for (auto&& pass : passes_) world().ILOG(" + {}", pass->name());
     world().debug_dump();
 
+    for (auto&& pass : passes_) pass->prepare();
```
NeuralCoder3, line 49: what is the reason to move the prepare before the `analyzed` and `curr_state` initialization?
The main difference I see is if the preparation introduces new externals.

NeuralCoder3 (approved):

thorin/pass/pass.cpp
fodinabor: yeah exactly, that's been the motivation here.
If externals need to be changed, this PR does so by creating a new nom and marking the old one as internal and the new one as external...

dialects/mem/phases/rw/add_mem.h
```diff
@@ -5,23 +5,21 @@
...
 private:
     const Def* add_mem_to_lams(Lam*, const Def*);
     const Def* rewrite_pi(const Pi*);
     const Def* mem_for_lam(Lam*) const;
 
-    Scheduler& sched() { return sched_.back(); }
-
-    std::vector<Scheduler> sched_;
-    Lam* curr_external_;
+    Scheduler sched_;
```
leissa, line 22: This is what I thought, that switching to `ScopePhase` we only need one `Scheduler`. However, I still think that - if we traverse the program properly - we don't need the scheduler at all. We can discuss that tomorrow.

NeuralCoder3:
One issue I found were in non-function objects with function types:
```
⊥:.Cn .Idx 4294967296
```
is not rewritten but should become
```
⊥:.Cn [%mem.M, .Idx 4294967296]
``` 

The same holds for top.
More generally, all types at the expression level (e.g. higher-order dependent axioms) should probably be rewritten.

leissa:
Looks good to me. One question though: Why do we need `AddMemWrapper`? Can't we just put `AddMem` and `PassManPhase` in this order into a `Pipeline`?

fodinabor:
Yes, that is the goal. Thus far, the only way to add the phase from within the plugin system seemed to be adding a Pass wrapper, that can be added to the extension points.
I would assume #144 might make this superfluous.

NeuralCoder3:
> Yes, that is the goal. Thus far, the only way to add the phase from within the plugin system seemed to be adding a Pass wrapper, that can be added to the extension points.

In between, there were also (not merged) extension points for phases.
But the correct change will come with #144.

NeuralCoder3:
Yes, this seems to solve this issue even without forcing optimizations to remove the offending code.

There seems to be an issue (before and after) with empty arguments
```
:4294967295: error: cannot pass argument 
  '(_1799704, tup_pb)' of type 
  '[%mem.M, .Cn [%mem.M, [], .Cn [%mem.M, .Idx 4294967296, .Idx 4294967296]]]' to 
  '(aug_pow_else_1799738, aug_pow_then_1800406)#_1800528' of domain 
  '[%mem.M, .Cn [%mem.M, .Cn [%mem.M, .Idx 4294967296, .Idx 4294967296]]]'
```

This error occurs during the add mem phase but might also be reshape related.

leissa:
Is this a show stopper or can we merge?

NeuralCoder3:
> Is this a show stopper or can we merge?

We can merge. This seems unrelated.

PR #150, opened 2022-11-25
Bugfix/axiom curry

leissa:
For a curried App of an Axiom, you only want to trigger normalization at specific spots. For this reason, Thorin maintains a Def::curry_ counter that each App decrements. The Axiom::normalizer() will be triggered when Axiom::curry() becomes `0`. These are also the spots that you can match/force. After that, the counter will be set to Axiom::trip(). E.g., let's say an Axiom has this type:
```
A -> B -> C -> D -> E
          ^         |
          |         |
          +---------+
```
Using an initial value as `5` for Axiom::curry and `3` as Axiom::trip has the effect that here
```
x a b c1 d1 e1 c2 d2 e2 c3 d3 e3
```
the Axiom::normalizer will be triggered after App'ing `e1`, `e2`, and `e3`

In the above case you can use Axiom::infer_curry_and_trip to automatically retrieve these values (happens as default in the frontend). You can specify other values, as you like in the frontend after the normalizer name:
```
.ax foo: A -> B -> C -> D -> E, normalize_foo, 3;    // curry 3, trip 0
.ax foo: A -> B -> C -> D -> E, normalize_foo, 3, 1; // curry 3, trip 1
```
See also the unit test `Axiom::curry` for details.

### DIFF
--- a/dialects/autodiff/auxiliary/autodiff_aux.cpp
+++ b/dialects/autodiff/auxiliary/autodiff_aux.cpp
@@ -146,7 +146,7 @@ const Def* zero_def(const Def* T) {
 const Def* op_sum(const Def* T, DefArray defs) {
     // TODO: assert all are of type T
     auto& world = T->world();
-    return world.raw_app(world.raw_app(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
+    return world.app<false>(world.app<false>(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
 }
 
 } // namespace thorin::autodiff
--- a/dialects/autodiff/auxiliary/autodiff_aux.cpp
+++ b/dialects/autodiff/auxiliary/autodiff_aux.cpp
@@ -146,7 +146,7 @@ const Def* zero_def(const Def* T) {
 const Def* op_sum(const Def* T, DefArray defs) {
     // TODO: assert all are of type T
     auto& world = T->world();
-    return world.app<false>(world.app<false>(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
+    return world.app(world.app(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
 }
 
 } // namespace thorin::autodiff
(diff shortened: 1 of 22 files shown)


dialects/autodiff/auxiliary/autodiff_aux.cpp
```diff
@@ -146,7 +146,7 @@ const Def* zero_def(const Def* T) {
 const Def* op_sum(const Def* T, DefArray defs) {
     // TODO: assert all are of type T
     auto& world = T->world();
-    return world.raw_app(world.raw_app(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
+    return world.app<false>(world.app<false>(world.ax<sum>(), {world.lit_nat(defs.size()), T}), world.tuple(defs));
```
NeuralCoder3, line 149: We want to normalize at this point.
The sum should be resolved to a chain of additions.

I think that most of the following `<false>` should be `<true>` in the general spirit of "always apply normalizers" (except the normalizer pass-through cases).

leissa: So, you are telling me, the original code should have used `app` instead of `raw_app`?

NeuralCoder3: Probably. I think I misunderstood `raw_app` and some of my code is broken/does not work as intended.

leissa:
@NeuralCoder3 I re-introduced the `raw_app` (now receiving the `type` of the `App`) because we don't need to recompute the dependent type twice. I also changed a couple of `raw_app`s in autodiff with proper `app`s. Seems to work.

I've also merged your latest work with inline optimizations into master and now in to this branch. Can you have a look whether this magic here actually fixes your problems?

NeuralCoder3:
> I've also merged your latest work with inline optimizations into master and now in to this branch. Can you have a look whether this magic here actually fixes your problems?

Yes, the normalization for recursive types and higher-order returns are now working.

dialects/direct/direct.thorin
```diff
@@ -22,7 +22,7 @@
 /// This axiom lets the user call a cps function in direct style.
 /// The function is not converted. Only the call site is changed.
 /// 
-.ax %direct.cps2ds: Π [T: *, U: *] -> (.Cn [T, .Cn U]) -> (T -> U), normalize_cps2ds;
+.ax %direct.cps2ds: Π [T: *, U: *] -> (.Cn [T, .Cn U]) -> (T -> U), normalize_cps2ds, 2;
```
leissa, line 25: Nice :)

PR #156, opened 2022-12-05
fixes #155

### LINKED ISSUE #155: Rewrite Phase Assertion Issue
There might be a bug in the rewrite phase or in my usage thereof.

In [277d500](https://github.com/NeuralCoder3/thorin2/commit/277d50087e082ce5aad7373fc0e81437ac942e08) in 
https://github.com/NeuralCoder3/thorin2/blob/matrix_dialect/dialects/matrix/passes/lower_matrix_lowlevel.cpp#L74

we have a minimal rewrite phase that just rewrites with the identity.

But for instance `./build/bin/thorin -o - -VVVV -d matrix lit/matrix/init_const_no_ret.thorin` fails at the assumption that
the world of the rewritten def or the gid is different.

leissa:
Bug fix for `Axiom::rebuild` - see #155.


fodinabor (approved):
LGTM!

PR #181, opened 2023-02-01
Higher order codegen

NeuralCoder3:
This branch contains fixes for higher-order code generation.
The fixes regard the llvm backend, memory dialect, and closure conversion dialect.

This branch is the base of the current state of automatic differentiation and the matrix dialect.

## Summary

Miscellaneous:
- .gitignore: ignore build folder but not everything with build in its name
- error.sh: small script to make it easier to provide issue information for failing tests

### Core
- dump filters for functions
- print list of executed passes
- fix lam_spec divergence

### Dialects
- refly: 
  - debug_dump: a pass that just dumps the world to stdout
- affine:
  - correctly propagate accumulator
- compile/opt:
  - move dialect conditionals from opt to compile for accessibility
  - generalize internal cleanup phase to avoid code duplication (parametrized over prefix)
- direct
  - fix issue of nested cps2ds calls using a stack
- mem
  - more mem manipulation operations (not assuming mem at position 1)
  - reshape: rewrite argument shape of a function (take special care of mems)
  - add_mem: documentation improvement, ignore argument position mems
- core:
  - add free
- clos:
  - move phase2pass wrapper 

### Tests
- allow filter annotation in printed thorin code
- more direct, mem, clos tests regarding mem issues


NeuralCoder3:
Note: The currently failing tests are due to the merge.

dialects/direct/passes/cps2ds.h
```diff
@@ -12,14 +14,18 @@ class CPS2DS : public RWPass<CPS2DS, Lam> {
 public:
     CPS2DS(PassMan& man)
         : RWPass(man, "cps2ds")
-        , cps_call_{world().sym("csp_call")} {}
+        , sym_{.cps_call = world().sym("csp_call")} {}
```
NeuralCoder3, line 17: fix spelling

dialects/mem/passes/rw/reshape.cpp
```diff
@@ -0,0 +1,337 @@
...
+    if (auto i = old2new_.find(def); i != old2new_.end()) return i->second;
+    auto new_def  = rewrite_def_(def);
+    old2new_[def] = new_def;
+    return new_def;
+}
+
+bool should_flatten(const Def* T) {
+    // handle [] cases
+    if (T->isa<Sigma>()) return true;
+    // also handle normalized tuple-arrays ((a:I32,b:I32) : <<2;I32>>)
```
NeuralCoder3, line 27: go over comments

dialects/core/be/ll/ll.cpp
```diff
@@ -790,6 +813,26 @@ std::string Emitter::emit_bb(BB& bb, const Def* def) {
         auto t      = convert(arith->type());
         auto mode   = as_lit(arith->decurry()->arg());
 
+        // #    if 0
```
NeuralCoder3, line 816: remove

thorin/pass/pipelinebuilder.h
```diff
@@ -22,6 +22,7 @@ class PipelineBuilder {
         auto pass = (Pass*)man->add<P>(std::forward<Args>(args)...);
         remember_pass_instance(pass, def);
     }
+    // TODO: add remembered entry
```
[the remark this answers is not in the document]
NeuralCoder3: Leave for later. For now, only pass instances are relevant.

dialects/core/be/ll/ll.cpp
```diff
@@ -313,7 +305,27 @@ void Emitter::emit_epilogue(Lam* lam) {
             }
         }
     } else if (auto ex = app->callee()->isa<Extract>(); ex && app->callee_type()->is_basicblock()) {
-        emit_unsafe(app->arg());
+        // A call to an extract like constructed for conditionals (else,then)#cond (args)
+        // TODO: we can not rely on the structure of the extract (it might be a nested extract)
```
leissa, line 309: If this is not the case, we should simply abort compilation; in addiont, we shouuld have a phase that splits nested extracts into basic blocks so the LLVM (or any other backend) doesn't have to fight this issue.

dialects/direct/passes/cps2ds.cpp
```diff
@@ -88,15 +113,36 @@ const Def* CPS2DS::rewrite_body_(const Def* def) {
...
+                            if (num.empty()) {
+                                new_name += "_1";
+                            } else {
+                                num      = num.substr(1);
+                                num      = std::to_string(std::stoi(num) + 1);
+                                new_name = new_name.substr(0, pos + append.size()) + "_" + num;
+                            }
+                        } else {
+                            new_name += append;
+                        }
```
leissa, line 133: I think this is a useful idiom that we want to factor out into a utility function. But for now, we can keep it like this.

dialects/mem/mem.h
```diff
@@ -105,5 +105,83 @@ inline Ref op_mslot(Ref type, Ref mem, Ref id) {
...
+
+    return arg;
+}
+
+/// Removes recusively all occurences of mem from a type (sigma).
+static Ref strip_mem_ty(Ref def) {
+    auto& world = def->world();
+
+    if (auto sigma = def->isa<Sigma>()) {
+        DefVec newOps;
```
leissa, line 142: `new_ops`

dialects/mem/mem.h
```diff
@@ -105,5 +105,83 @@ inline Ref op_mslot(Ref type, Ref mem, Ref id) {
...
+        for (auto op : sigma->ops()) {
+            auto newOp = strip_mem_ty(op);
+            if (newOp != world.sigma()) { newOps.push_back(newOp); }
```
leissa, line 145:
```c++
if (auto new_op = strip_mem_ty(op); new_op != world.sigma()) new_ops.push_back(new_op);
```

dialects/mem/mem.h
```diff
@@ -105,5 +105,83 @@ inline Ref op_mslot(Ref type, Ref mem, Ref id) {
...
+/// Returns an empty tuple if applied with mem alone.
+static Ref strip_mem(Ref def) {
+    auto& world = def->world();
+
+    if (auto tuple = def->isa<Tuple>()) {
+        DefVec newOps;
+        for (auto op : tuple->ops()) {
+            auto newOp = strip_mem(op);
+            if (newOp != world.tuple()) { newOps.push_back(newOp); }
+        }
```
leissa, line 166: see above

dialects/mem/mem.h
```diff
@@ -105,5 +105,83 @@ inline Ref op_mslot(Ref type, Ref mem, Ref id) {
...
+        return world.tuple();
+    } else if (auto extract = def->isa<Extract>()) {
+        // The case that this one element is a mem and should return () is handled above.
+        if (extract->num_projs() == 1) { return extract; }
+
+        DefVec newOps;
+        for (auto op : extract->projs()) {
+            auto newOp = strip_mem(op);
+            if (newOp != world.tuple()) { newOps.push_back(newOp); }
+        }
```
leissa, line 179: see above

lit/clos/lib.c
```diff
@@ -1,13 +1,54 @@
...
-void print_i32(int32_t i) { printf("%" PRId32 "\n", i); }
+// #define SILENT 1
+
+#ifdef SILENT
+// less intrusive than redefining printf
+void print_i32(int32_t i) { }
+void println_i32(int32_t i) { }
+void newline() { }
 
+void printInteger(int i) {}
```
leissa, line 19:
```c
print_int
```

lit/clos/lib.c
```diff
@@ -1,13 +1,54 @@
...
+void printIntegerNL(int i) {}
```
leissa, line 20:
```c
print_int_nl
```

lit/clos/lib.c
```diff
@@ -1,13 +1,54 @@
...
+void printNL() {}
```
leissa, line 21:
```c
print_nl
```

lit/mem/arg_style_no_extern.thorin.disabled
```diff
@@ -0,0 +1,176 @@
+// RUN: rm -f %t.ll
```
leissa, line 1: eh, what is this test doing? Oo

dialects/core/be/ll/ll.cpp
NeuralCoder3: I opened #187 and reverted the change.

lit/mem/arg_style_no_extern.thorin.disabled
NeuralCoder3: Are you referring to the missing ` ; \` after `rm -f %t.ll`?

Currently, the test does nothing as it is disabled (issue with MacOS).
But in theory, it enumerates different possibilities of argument shapes (flat, arg, mem),
selects the current one using the c preprocessor, and calls it.

The automatic shape check is not yet implemented.

NeuralCoder3:
> On a more stylistic note, I think we should remove the space just in front of the `@` when dumping filters. Say, we have a curried function:
> 
> ```
> .con foo(a: T)@(.tt)(b: U) = /*...*/
> ```
> 
> With the current logic, it would look like this:
> 
> ```
> .con foo(a: T) @(.tt)(b: U) = /*...*/
> ```
> 
> And now it looks as if the filter belongs to the **second** curried function.

Do curried functions still exist at the dumping stage?
I would expect them to be syntactic sugar that is resolved at parsing.

leissa:
> > On a more stylistic note, I think we should remove the space just in front of the `@` when dumping filters. Say, we have a curried function:
> > ```
> > .con foo(a: T)@(.tt)(b: U) = /*...*/
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With the current logic, it would look like this:
> > ```
> > .con foo(a: T) @(.tt)(b: U) = /*...*/
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And now it looks as if the filter belongs to the **second** curried function.
> 
> Do curried functions still exist at the dumping stage? I would expect them to be syntactic sugar that is resolved at parsing.

correct. But I still find it better for the sake of consistency to remove the space.

lit/mem/arg_style_no_extern.thorin.disabled
leissa: > But in theory, it enumerates different possibilities of argument shapes (flat, arg, mem), selects the current one using the c preprocessor, and calls it.

That was the answer to my question :)

PR #183, opened 2023-02-20
removed doxygen workaround

leissa:
* has been fixed upstream; see doxygen/doxygen#9668


leissa:
We need to wait for an official release in order to integrate this into our doxygen runner.

PR #184, opened 2023-02-25
Feature/strpool

leissa:
Migrating `const Def* dbg` to `Loc`, `Sym`, and `const Def* meta`.


dialects/affine/passes/lower_for.h
```diff
@@ -9,11 +9,22 @@ namespace thorin::affine {
 class LowerFor : public RWPass<LowerFor, Lam> {
 public:
     LowerFor(PassMan& man)
-        : RWPass(man, "lower_affine_for") {}
+        : RWPass(man, "lower_affine_for")
```
NeuralCoder3, line 12: We can leave this in as an example.
But as you write above, the speedup is not worth the effort, indirection, redundancy, and maintaining overhead.
(This comment is with regard to the sym_ construction)

dialects/autodiff/auxiliary/autodiff_rewrite_toplevel.cpp
```diff
@@ -10,7 +10,8 @@ const Def* AutoDiffEval::derive_(const Def* def) {
     auto lam    = def->as_nom<Lam>(); // TODO check if nominal
     world.DLOG("Derive lambda: {}", def);
     auto deriv_ty = autodiff_type_fun_pi(lam->type());
-    auto deriv    = world.nom_lam(deriv_ty, world.dbg(lam->name() + "_deriv"));
+    auto deriv    = world.nom_lam(deriv_ty)->set(lam->sym());
+    deriv->debug_suffix("_deriv");
```
NeuralCoder3, line 14: If I read the code correctly, this suffix is only added if the compilation runs in debug mode.
However, we always need the suffix.

Example:
```
x = pow(2,3)
y,y_pb = ad pow (2,3)
```
becomes
```
x = pow(2,3)
y,y_pb = pow_diff (2,3)
```

We need both versions (function and AD'ed function) as separate entities.

dialects/clos/phase/clos_conv.cpp
```diff
@@ -284,19 +283,21 @@ ClosConv::Stub ClosConv::make_stub(const DefSet& fvs, Lam* old_lam, Def2Def& sub
     auto num_fvs     = fvs.size();
     auto env_type    = rewrite(env->type(), subst);
     auto new_fn_type = type_clos(old_lam->type(), subst, env_type)->as<Pi>();
-    auto new_lam     = old_lam->stub(w, new_fn_type, w.dbg(old_lam->name()));
-    new_lam->set_debug_name((old_lam->is_external() || !old_lam->is_set()) ? "cc_" + old_lam->name() : old_lam->name());
+    auto new_lam     = old_lam->stub(w, new_fn_type);
+    // TODO
+    // new_lam->set_debug_name((old_lam->is_external() || !old_lam->is_set()) ? "cc_" + old_lam->name() :
+    // old_lam->name());
```
NeuralCoder3, line 289: Why not just set the name as before?
Is there a problem with the name or is it simple not necessary?

dialects/direct/passes/cps2ds.cpp
```diff
@@ -128,9 +129,9 @@ const Def* CPS2DS::rewrite_body_(const Def* def) {
 
     if (auto old_nom = def->isa_nom()) { return old_nom; }
     DefArray new_ops{def->ops(), [&](const Def* op) { return rewrite_body(op); }};
-    if (def->isa<Tuple>()) return world.tuple(new_ops, def->dbg());
+    if (def->isa<Tuple>()) return world.tuple(new_ops); // TODO just remove this line? there is rebuild below
```
NeuralCoder3, line 132: There were problems where the tuple would be rebuilt but the old type `def->type()` was (wrongly) assigned to the new tuple.
 If this issue/behaviour is fixed, the line can be removed after a test.

dialects/affine/passes/lower_for.h
leissa: yeah, I think I will just remove this again

dialects/autodiff/auxiliary/autodiff_rewrite_toplevel.cpp
leissa: ic. can you grep for `debug_suffix` and `debug_prefix` whether I did the same mistake somewhere else?

dialects/direct/passes/cps2ds.cpp
leissa: This line is already gone in your `ho_codegen` branch, so we can leave it like this for now.

dialects/autodiff/auxiliary/autodiff_rewrite_toplevel.cpp
NeuralCoder3: I do not know if eta_exp has a problem with it.

In cps2ds, ds2cps, the `_cont` suffix is necessary. 
The `_cps` suffix is not strictly necessary (the original function should not be referenced) but the functions are only resolved in another step. They are also not semantically equivalent (another type scheme).
Therefore, I would always add the suffix.

leissa: alright. the name *should* not really be part of the semantics, but we had this discussion that right now, we don't have a better mechanism, so in your case it makes sense. The use of `debug_*fix` in ssa_constr and eta_exp is intentional.

Issue #171, opened 2023-01-12
Automatic Dialect Loading


NeuralCoder3:
We probably should load dialects automatically.
Currently, the user has to specify the loaded dialects using `-d [dialect]`.

A fine control might be helpful during development (for instance ignoring closure conversion phases).
But we can handle these settings (disabling some phases) also using the opt primitives.
Additionally, it is quite easy to forget to load a dialect and wonder why some passes are not executed (see #169).

One option would be to always load all available phases.

Another would be to load all dialects imported in the file.
This however misses dialects that are recursively needed (e.g. maths in matrix).
If we recursively look up imported dialects, we end up again loading all of them as opt imports nearly every other dialect.

leissa:
Maybe a `--bootstrap` option is enough to not load the plugins and only parse the `*.thorin` files. If `--bootstrap` is not present all plugins should be auto-loaded.

NeuralCoder3:
We have something similar already:
If not calling with `--output-h`, we automatically load `core`, `mem`, `compile`, `opt`.
So the only difference would probably be to remove the `-d` option and to search for all available dialects instead of this hardcoded list.

leissa:
Right. `--output-h` implies "bootstrap" mode.

Issue #143, opened 2022-11-11
optimization exhibits non-deterministic behavior


NeuralCoder3:
Sometimes, the behavior of the optimization pipeline seems to be non-deterministic.

Example:
./build/bin/thorin -d mem -o - lit/mem/no_mem.thorin -VVVV
in
https://github.com/NeuralCoder3/thorin2/tree/ad_ptr_merge
702d848

The issue might be due to the add_mem optimization, the pipeline builder, or an underlying bug in thorin.

leissa:
Yes. this is super annoying. Another source is this:
```c++
world.app(emit1(), emit2());
```
It's implementation defined whether `emit1()` is happened first or second. This code has different behavior on different compilers/OS's.

I have implemented the `--trace-gids` switch that we could somehow use to test for this in our CI.

NeuralCoder3:
The issue happens only sometimes on with the same executable on the same computer in the same cirumstances.
Therefore, timing issues or randomness might be the cause.
 
Probably related issue:
`./build/bin/thorin -d matrix -d affine lit/matrix/mapReduce_mult.thorin -o - -VVVV` in `matrix_dialect` [f3a3def](https://github.com/NeuralCoder3/thorin2/commit/f3a3def5b52d472db30718f60b27ffa77c44f948)
sometimes generates thorin code and sometimes prints the following error:
```
:4294967295: error: cannot pass argument 
  '(__806508#2:(.Idx 3), ‹__806508#2:(.Idx 3); .Idx 4294967296›, 0)' of type 
  '[.Nat, «__806508#2:(.Idx 3); ★», .Nat]' to 
  '%mem.lea' of domain 
  '[n_834521: .Nat, _834535: «n_836768; ★», _834540: .Nat]'
``` 

which seems odd to me as the arguments are of the style
``` 
(n, <n; T>; 0)
```
which should be the type
```
[n:.Nat, <<n; *>>; .Nat]
```
which should agree with lea.

leissa:
Was fighting this issue in #184 as a Debug build produced different outputs as the Release one
* 05e833b23e3318b1441f3548cbf8d636b6f0502b 
    A few asserts created new Defs resulting in slightly different behavior between Debug and Release builds. This commit fixes the issue.
* 2997a1d9f7a611b5aa48cea33a7b008ded00e99d
    This one fixes a subtle problem when a `Def`'s name is coincidentally the same as for an external `Def`.
    
[...]

Issue #116, opened 2022-10-14
Import of files


NeuralCoder3:
Currently, the import mechanism operates on dialects.

One could imagine a more modular style of files with a file-based import.
 
```rust
.import name // imports dialect name
.import name/abc // import abc.thorin from the dialect provided files
.import path/abc // imports path/abc.thorin from a local file
```

Local imports would allows for standard definitions like
```
.let _32 =  4294967296;
.let I32 = .Idx _32;
.let r32 = %core.Real 32;
.let double = %core.Real 64;
```

Dialect imports would allow for a more readable and modular approach to dialect definitions.
One example would be the differentiated constructs of `autodiff`.

leissa:
While #185 fixes the underlying problem, I'll keep this one open until we have worked out a proper solution who we want to deal with imports and scoping.

leissa:
While a more sophisticated scoping/import system is still not there yet a basic imports work. A better system is sth for a future version of Mim.

Issue #99, opened 2022-10-10
Externals


NeuralCoder3:
Dialects currently can not declare externals.
This is probably due to multiple reads of the dialect thorin file.

This behavior can be observed in the automatic differentiation branch.

The intermediate fix is to disable the insertion/containment check in `world.h`.

PR #185, opened 2023-03-08
Bugfix/parse once

leissa:
# Fixes:

* #99 
* #116
    For now, the `XXX` within `.import XXX` must be a single name in your search paths
* #143 
    There were multiple reads of some files that sometimes caused funny behavior.
* #171
    You can now load a plugin using `.plugin XXX;` in your `.thorin` file.
    
# Changes

* All the bits and pieces which have been passed around manually to make the plugin architecture work are now streamlined and centrally manged in `Driver`.
* File names & paths now use `fs::path` - including `Loc`.
* The c++ sources now use the term `plugin` instead of `dialect`
* more migration from `const Def*` -> `Ref`
* a little bit more of Infer magic
* Wrapper `assert_emplace(container, val)` to assert that an insertion actually took place
* Wrapper `lookup(map, key)` that returns a pointer to value if found or `nullptr` otherwise.

# Remarks

Although we can define now all kind things in our `imports`, I'm reluctant to actually use this feature because right now everything lives in a global namespace. We probably want to have sth smarter.


NeuralCoder3:
For clarification:

plugin
=> make passes normalizers ... available
import 
=> makes axioms, local definitions, ... in thorin available

Is this correct?

And the relation would be:
plugin -> import
but not import -> plugin
.

NeuralCoder3:
> Although we can define now all kind things in our imports, I'm reluctant to actually use this feature because right now everything lives in a global namespace. We probably want to have sth smarter.

Prefixes like `%name.name` are reserved for axioms (we can not prefix other declarations/definitions), right?

NeuralCoder3:
Is it currently just not necessary that dialects load other dialects using `plugin` or should this be avoided in general?

thorin/dump.cpp
```diff
@@ -382,7 +382,7 @@ void World::dump(std::ostream& os) {
         auto dep    = DepTree(*this);
         auto dumper = Dumper(os, &dep);
 
-        for (const auto& import : imported()) print(os, ".import {};\n", import);
+        for (auto [_, name] : driver().imports()) print(os, ".import {};\n", name);
```
NeuralCoder3, line 385: If possible, it would probably be nice to also dump loaded plugins as `.plugin`.

cli/main.cpp
```diff
@@ -45,14 +44,14 @@ int main(int argc, char** argv) {
         auto cli = lyra::cli()
             | lyra::help(show_help)
             | lyra::opt(show_version             )["-v"]["--version"           ]("Display version info and exit.")
-            | lyra::opt(list_dialect_paths       )["-l"]["--list-dialect-paths"]("List search paths in order and exit.")
+            | lyra::opt(list_search_paths        )["-l"]["--list-search-paths" ]("List search paths in order and exit.")
             | lyra::opt(clang,          "clang"  )["-c"]["--clang"             ]("Path to clang executable (default: '" THORIN_WHICH " clang').")
-            | lyra::opt(plugins,        "dialect")["-d"]["--dialect"           ]("Dynamically load dialect [WIP].")
-            | lyra::opt(dialect_paths,  "path"   )["-D"]["--dialect-path"      ]("Path to search dialects in.")
+            | lyra::opt(plugins,        "dialect")["-d"]["--dialect"           ]("Dynamically load plugin.")
```
NeuralCoder3, line 49: If dialects are now called plugins, the option should probably be
`-p` `--plugin`
and
`-P` `--plugin-path`.

leissa:
> > Although we can define now all kind things in our imports, I'm reluctant to actually use this feature because right now everything lives in a global namespace. We probably want to have sth smarter.
> 
> Prefixes like `%name.name` are reserved for axioms (we can not prefix other declarations/definitions), right?

correct. I'd like to change this but this is sth for another PR.

cmake/Thorin.cmake
```diff
@@ -101,7 +101,7 @@ function(add_thorin_dialect)
         OUTPUT ${DIALECT_MD} ${DIALECT_H}
```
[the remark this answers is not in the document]
leissa: yes, this is why I'm gradually doing this renaming. I think just renaming files isn't a big deal. Git should be able to cope with it. Only the include section may get messed up in a merge. But again, more PRs are following.

cli/main.cpp
leissa: yes. Also thought about this, but for now I didn't want to touch each and every test case

thorin/dump.cpp
leissa: good point. I will do that.

PR #191, opened 2023-03-27
Refactor/plugin

leissa:
More migration dialect -> plugin

* Everything except the directory structure now uses the term "plugin" instead of "dialect"
* CLI: 
    * `-d` -> `-p`
    * `-D` -> `-P`
    * make bootstrapping explicit via `--bootstrap`
* Migrating lit tests to use `.plugin` instead of `-p` if possible


lit/autodiff/autodiff_mult_in_call2.thorin
```diff
@@ -1,9 +1,8 @@
 // RUN: rm -f %t.ll ; \
-// RUN: %thorin -d direct -d autodiff %s --output-ll %t.ll -o - | FileCheck %s
+// RUN: %thorin -p direct %s --output-ll %t.ll -o - | FileCheck %s
```
fodinabor, line 2: To my understanding using `.plugin direct` should have the same effect.. so why use `-d direct` here?

leissa: My line of thinking was that we probably want to keep the option to load different plugins via the command line to change the compilation behavior of the program. Sth like:
```
thorin foo.thorin -p O2 -p ll
```
(and remove the current builtin `-O2` and `--output-ll` switches)
Now, this particular test case doesn't explicitly use any axiom of the `direct` plugin. For this reason, I'm only specifying it via the CLI. But yes, it's debatable whether we want to use the CLI or `.plugin direct` here.

PR #193, opened 2023-03-28
Refactor/core

leissa:
# Changes

* `%core.nop` -> `%core.nat`: 
    `nop` usually means sth entirely differently.
* Use **implicit** `D`st type for `%core.bitcast`:
    `.ax %core.bitcast: Π [D: *, S: *][S] -> D` -> `.ax %core.bitcast: Π.[S: *][D: *][S] -> D`
* showcasing some of the newer features of Thorin in `lit/clos/malloc.thorin`
* adds wrap around parameter to deal with non-power of two `.Idx` in `core::bit1` & `core::bit2`
    See Discord discussion. Instead of casting to `bit_floor` or adding proof certificates, I just added the solution we already have for `%core::wrap`: add another (`.Nat`) flag that determines the overflow behavior. Rn, however, nobody is really doing anything with this flag. This is sth for future work.

### DIFF
--- a/dialects/CMakeLists.txt
+++ b/dialects/CMakeLists.txt
@@ -92,6 +92,8 @@ add_thorin_plugin(core
     DEPENDS
         math
         mem
+    HEADER_DEPENDS
+        clos
     INSTALL
 )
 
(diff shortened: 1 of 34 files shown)


NeuralCoder3:
> Use implicit Dst type for %core.bitcast:

The source type is now implicit not the destination type, right?
(The dst type could theoretically be implicit but could not be locally inferred; the information is only available in its usage context (Hindley–Milner style inference/unification))

NeuralCoder3 (approved):

leissa:
> > Use implicit Dst type for %core.bitcast:
> 
> The source type is now implicit not the destination type, right? (The dst type could theoretically be implicit but could not be locally inferred; the information is only available in its usage context (Hindley–Milner style inference/unification))

correct.

PR #195, opened 2023-03-31
Refactor/mut

leissa:
As I've already pointed out several times the name *nominal* is confusing as *nominals* are in fact **not** [nominal](https://en.wikipedia.org/wiki/Nominal_type_system) as they are tested like structurals for alpha-equiv.

This patch renames:
* **nom**inal -> **mut**able
* **structural** -> **imm**utable

In the future I'd like to add the *option* to make **mut**ables truely *nominal*/unique.

In addition: Some more porting from `const Def*` -> `Ref`.

### DIFF
--- a/dialects/affine/passes/lower_for.cpp
+++ b/dialects/affine/passes/lower_for.cpp
@@ -12,18 +12,18 @@ const Def* LowerFor::rewrite(const Def* def) {
 
     if (auto for_ax = match<affine::For>(def)) {
         auto& w = world();
-        w.DLOG("rewriting for axiom: {} within {}", for_ax, curr_nom());
+        w.DLOG("rewriting for axiom: {} within {}", for_ax, curr_mut());
 
         auto for_pi = for_ax->callee_type();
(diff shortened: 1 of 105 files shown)


NeuralCoder3:
> In addition: Some more porting from const Def* -> Ref.

Is there an easy rule for when to use `Ref` and when to use `Def` (`const Def*`)?
E.g. should a plugin developer 
- always use `const Def*`
- always use `Ref`
- use `const Def*` except for normalizers?

NeuralCoder3 (approved):

leissa:
> > In addition: Some more porting from const Def* -> Ref.
> 
> Is there an easy rule for when to use `Ref` and when to use `Def` (`const Def*`)? E.g. should a plugin developer
> 
>     * always use `const Def*`
> 
>     * always use `Ref`
> 
>     * use `const Def*` except for normalizers?

The rule is to always use `Ref` instead of `const Def*`. For everything else such as `Def*` or `const App*` or `Sigma*`, use these types instead.

PR #200, opened 2023-04-14
Cleanup

leissa:
* `refly::debug_print` is now internal
* `autodiff_aux.*` integrated into `autodiff.*`
* `continuation_dom`/`codom` -> `Lam`/`Pi::cn_dom`/`codom`
* removed `f_arg_ty` macro
* improved docs


NeuralCoder3:
When is a method `isa_` and when `is_`?

dialects/mem/passes/rw/reshape.cpp
```diff
@@ -40,6 +37,45 @@ bool should_flatten(const Def* T) {
...
+    } else {
+        types.push_back(T);
+    }
+    return types;
+}
+
+// TODO try to remove code duplication with flatten_ty
+std::vector<const Def*> flatten_def(const Def* def) {
+    std::vector<const Def*> defs;
+    if (should_flatten(def->type())) {
```
NeuralCoder3, line 57: The merge with flatten_ty could just be
`is_type ? def : def->type()`

thorin/def.cpp
```diff
@@ -73,107 +75,162 @@ UMax::UMax(World& world, Defs ops)
...
+template TBound<true >* TBound<true >::stub(World&, Ref);
+template TExt<false>*   TExt<false>  ::stub(World&, Ref);
+template TExt<true >*   TExt<true >  ::stub(World&, Ref);
+#endif
+
 // clang-format on
 
 /*
- * restructure
+ * freeze
```
NeuralCoder3, line 152: `freeze` constructs an immutable (formerly structural) from mutable (formerly nominal) by rewrapping its arguments and possibly simplifying it along the way?

NeuralCoder3 (approved):

thorin/def.cpp
leissa: It *tries* to make an immutable from a mutable. This usually works if the mutable isn't recursive and its var isn't used.
Will add it to the docs.

leissa:
> When is a method `isa_` and when `is_`?

See [here](https://anydsl.github.io/thorin2/coding.html#autotoc_md2):
> * methods/functions that return a `bool` should be prefixed with `is_`
> * methods/functions that return a `std::optional` or a pointer that may be `nullptr` should be prefixed with `isa_`

thorin/def.cpp
leissa: `freeze` was a bit cryptic. It's now called `immutabilize`.

PR #204, opened 2023-04-20
Rang

leissa:
Use [rang](https://github.com/agauniyal/rang) for colored output.


thorin/util/log.h
```diff
@@ -42,8 +43,8 @@ class Log {
     void log(Level level, Loc loc, const char* fmt, Args&&... args) const {
         if (ostream_ && level <= max_level_) {
             std::ostringstream oss;
-            oss << loc;
-            print(ostream(), "{}:{}: ", colorize(level2acro(level), level2color(level)), colorize(oss.str(), 7));
+            print(ostream(), "{}{}:{}{}:{} ", level2color(level), level2acro(level), rang::fg::gray, loc,
```
NeuralCoder3, line 46: From the code, I guess that something (probably only the acronym?) should be colorful.

In my test (e.g. `./build/bin/thorin lit/demo/const.thorin -o - -VVVV`) without color anywhere.

NeuralCoder3 (approved):

leissa:
Using Konsole with
```sh
$ echo $TERM  
xterm-256color
```

PR #207, opened 2023-04-26
Feature/fun

leissa:
This adds `.fun` decls, `.fn` exprs, and `.Fn` types.

You can now write decls like this:
```
.fun f (T: *)(a: T, b: T) -> T = return a;
```
which means the same as
```
.con f (T: *)((a: T, b: T), return: .Cn T) = return a;
```

There is an according expr available:
```
.let f = .fn (T: *)(a: T, b: T) -> T = return a;
```

And a type:
```
.let F = .Fn [T: *][T, T] -> T;
```

Continuation types can now also be curried:
```
.let K = .Cn [T: *][T, T];
```
Finally, you can invoke returning continuations with the ret expression:
```
.let res = f .Nat $ (23, 42);
use(res)
```

Includes already #206.


docs/langref.md
```diff
@@ -274,6 +284,51 @@ This has the effect that, e.g., `Π T: * → T → T` has the expected binding l
...
+```
+
+#### Expressions
+
+The following function *expressions* are all equivalent.
+What is more, since they are bound by a *let declaration*, they have the exact same effect as the function *declarations* above:
+```
+.let f = .lm (T: *)((x y: T), return: T -> ⊥) -> ⊥ = return x;
+.let f = .cn (T: *)((x y: T), return: .Cn T)       = return x;
+.let f = .fn (T: *) (x y: T)                       = return x;
```
NeuralCoder3, line 313: Same as above: return type is midding for `fn`.

docs/langref.md
```diff
@@ -274,6 +284,51 @@ This has the effect that, e.g., `Π T: * → T → T` has the expected binding l
...
+```
+f .Nat ((23, 42),.cn res: .Nat = use(res))
+.ret res = f .Nat : (23, 42); use(res)
+```
+#### Function Types
+
+Finally, the following function types are all equivalent and denote the type of `f` above.
+```
+ Π [T:*][T, T][T -> ⊥] -> ⊥
+.Cn[T:*][T, T][.Cn T]
```
NeuralCoder3, line 328: Shouldn't it be 
```
 Π [T:*][[T, T][T -> ⊥]] -> ⊥
.Cn[T:*][[T, T][.Cn T]]
```

If not spelled out explicitly, the text should make clear that `.Cn A` is syntactic sugar for `A -> .Bot`.

docs/langref.md
```diff
@@ -274,6 +284,51 @@ This has the effect that, e.g., `Π T: * → T → T` has the expected binding l
...
+.let f = .cn (T: *)((x y: T), return: .Cn T)       = return x;
+.let f = .fn (T: *) (x y: T)                       = return x;
+```
+
+#### Applications
+
+The following expressions for applying `f` are also equivalent:
+```
+f .Nat ((23, 42),.cn res: .Nat = use(res))
+.ret res = f .Nat : (23, 42); use(res)
```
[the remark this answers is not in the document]
leissa: It's `.ret name = f $ args`. Only the docs were wrong.

PR #209, opened 2023-04-27
Feature/sigma

leissa:


lit/sigma.thorin
```diff
@@ -0,0 +1,20 @@
...
+// RUN: %thorin %s -o -
+.plugin math;
+
+.let _64 = (52, 11);
+.let F64 = %math.F _64;
+
+.Sigma Vec1 = [x: F64];
+.Sigma vec1 = [X: F64];
+.Sigma Vec3: *, 3;
+.Sigma Vec3: *, 3 = [x: F64, y: F64, z: F64];
```
NeuralCoder3, line 11: If I understand the syntax correctly,
`.Sigma Vec3: *, 3;`
means:
``` 
(Coq Code)
Definition Vec3 : Type. (* + Additional information that it is isomorphic to Type^3 *)
``` 

`.Sigma Vec3: *, 3 = [x: F64, y: F64, z: F64];` 
means:
``` 
Record Vec3 : Type := _mkVec3 {
  x : F64, y: F64, z: F64
}.
```


Would 
```
.let Vec3 = [x: F64, y: F64, z: F64];
or
.let Vec3 = [ F64, F64, F64];
``` 
also be valid?

NeuralCoder3 (approved):

lit/sigma.thorin
leissa: Correct and yes. Difference is that the `.let` variant generates an *immutable* Sigma which also means that the `x`, `y`, `z` will be "forgotten" outside the sigma expression. So, this won't work:
```
.let Vec3 = [x: F64, y: F64, z: F64];
.let v: Vec3 = ...;
.let x = v#x; 
```
Also, the `.Sigma`-decl allows for recursion.

PR #210, opened 2023-04-28
Start working on a RegEx dialect as a normalization showcase.

fodinabor:
A simple use case that showcases the power of the normalization framework present in Thorin is regular expressions.
Defines a set of axioms that represent ranges, any character, alternatives, sequences and quantifiers.
These should be sufficient to define (most) compile-time regular expressions (missing `{n,m}` quantifiers, lookaheads, .. atm)

Accompanying the axioms are normalizers that simplify the RegEx.
An example of normalization is quantifier merging:
`(\d*)?`, `(\d?)*`, `(\d+)?`, `(\d?)+`, `(\d*)+`, `(\d+)*` all normalize to `\d*`.
For disjunctions: duplicates are removed, opposite classes (e.g. `[\s\S]`) are reduced to match any character "`.`".
All classes and literals are translated to disjunctions of ranges and the ranges are merged.
Disjunctions and conjunctions are always normalized to only have two arguments to make the matcher impl simpler.

There is also a phase that recursively replaces the applied regex with rather primitive matcher functions.
Therefore we can actually match the regex. Note, currently only deterministic regex can be matched -> `\w\w+` works, but `\w+\w` doesn't.

https://github.com/fodinabor/thorin_regex_benchmark contains a benchmark comparing Thorin's against [CTRE](https://github.com/hanickadot/compile-time-regular-expressions), [PCRE2](https://pcre.org) and `std::regex`:
```
engine,average[us],min[us],max[us],deviation[%],runs[us]
pcre2_jit,  1032,   1026,   1046,   1,  1033 1032 1041 1046 1037 1027 1026 1026 1028 1029 
ctre,       4053,   4049,   4065,   0,  4051 4059 4052 4051 4049 4050 4065 4051 4050 4051 
thorin,     909,    893,    1017,   13, 897 893 1017 902 901 894 897 894 899 896 
pcre2,      3028,   3016,   3071,   1,  3071 3029 3016 3019 3027 3026 3028 3025 3020 3024 
std,        9232,   8837,   10141,  14, 9204 10141 8837 9028 9186 9139 9153 9173 9167 9294 
```
Which looks pretty favorable for thorin on this regex benchmark.
Similarly, the compilation time of the thorin matcher is looking good (note, pcre2 performs runtime compilation, not measured by either of the metrics):
```
[...]

PR #211, opened 2023-05-02
Support for char & string literals

leissa:
Added support for char and string literals. 

* `'x'` is a literal of type `I8`.
* `"ABC\0"`, is a tuple `` (`A`, `B`, `C`, `\0`) `` of type `«4; I8»`.
* Note that Thorin doesn't care about null-termination or whatever. So in order to use null-terminated strings, you have to take care of `\0` by yourself.
* Check out `lit/str.thorin` as a test case/example.
* As `'x` doesn't work anymore as a rebind operator, this branch changes this to `` `x `` instead.
* Rn, this only adds ASCII support with the usual escape chars from C.

We probably want to use this for #210.


fodinabor (approved):
Thx, that will indeed be quite useful for #210!

PR #212, opened 2023-05-03
Feature/extensions

leissa:
# Changes

* The axiom namespace has been generalized to work with all kinds of declarations. <br>
    Examples:
    ```
    .let %math.f16     = (10,  5);
    .let %math.f32     = (23,  8);
    .let %math.f64     = (52, 11);
    .let %math.bf16    = ( 7,  8);
    .let %math.nvtf32  = (10,  8);
    .let %math.amdfp24 = (16,  7);
    .let %math.pxr24   = (15,  8);

    .let %math.F16     = %math.F %math.f16;
    .let %math.F32     = %math.F %math.f32;
    .let %math.F64     = %math.F %math.f64;
    .let %math.BF16    = %math.F %math.bf16;
    .let %math.NVTF32  = %math.F %math.nvtf32; // actually 19 bits; aligns to 32 bit
    .let %math.AMDFP24 = %math.F %math.amdfp24;
    .let %math.PXR24   = %math.F %math.pxr24;
    ```
    ```
    .lam %mem.Ptr0 T: * -> * = %mem.Ptr (T, 0);
    ```
* You can use `World::annex` and `World::call` to use these decls from within C++:
    ```c++
    auto i32_t = world.annex<core::I32>();
    auto ptr_t = world.call<mem::Ptr0>(i32_t);
    ````
* Decoupled a couple of things from `Axiom` into new `Annex` class
* Ported test cases to use new features
* Removed some now superfluous wrappers/glue code
* Some new test cases
* Fixed a few bugs
* Impala branch: [feature/extensions](https://github.com/AnyDSL/impala/tree/feature/extensions)
* Includes:
    * #207 
    * #209
   
## Some more Things

As the number of branches were running out of hand, I pushed a few things on this branch although they had not much to do with my original PR:

* `%mem.m`, `%mem.ignore`, and wrapper for both `%mem.rm` to ignore side effects
* `a @ b` expression to call `a` and manually provide `b` as argument that would normally be inferred
* check out test case `lit/mem/mem_rm.thorin`
* llvm bug fix for 1 tuples


NeuralCoder3:
`annex` now subsumes axioms and definitions?
Therefore, to create something, `annex` can be used.
To inspect something `axiom` is used (and definitions can not be inspected by their `annex` identifier)?

dialects/mem/mem.thorin
```diff
@@ -74,22 +76,34 @@
...
+/// Load effective address.
 /// Performs address computation by offsetting the passed pointer with index `i`.
 .ax %mem.lea: Π [n: .Nat, Ts: «n; *», as: .Nat][%mem.Ptr(«j: n; Ts#j», as), i: .Idx n] -> %mem.Ptr(Ts#i, as), normalize_lea;
 ///
+/// ### %mem.create/destroy
+///
+/// Create side effect out of thin or pretend that the side effect does not exists.
+/// @warning Use with caution since you completely remove the `%%mem.M` dependency.
+///
+.ax  %mem.m: %mem.M;
```
NeuralCoder3, line 88: This is automatically resolved in the already existing memory-polishing phases and as such equivalent to `.Top : %mem.M`?

thorin/world.h
```diff
@@ -149,21 +149,38 @@ class World {
...
-    const auto& axioms() const { return move_.axioms; }
+    const auto& annexes() const { return move_.annexes; }
     const auto& externals() const { return move_.externals; }
-    bool empty() { return move_.externals.empty(); }
-    void make_external(Def* def) {
-        assert(!def->external_);
-        def->external_ = true;
-        assert_emplace(move_.externals, def->sym(), def);
+    void make_external(Def* def, bool on) {
+        assert(!def->external_ == on);
```
NeuralCoder3, line 155: maybe rename to `toggle_external` or `set_external`?

leissa:
> > a @ b expression to call a and manually provide b as argument that would normally be inferred
> 
> `@` is an infix in this case? If I use `@`, I need to provide all implicit arguments, right? Maybe the Coq syntax `@a b` would be a bit cleaner.

`a @ b` is Infix und only applies to `one` argument. I also thought about `@a b`. Lean has the same feature. It's just, that it is more difficult to implement. But I will give it a shot. I also prefer the all-implicits-removed-prefix variant.

leissa:
> `annex` now subsumes axioms and definitions? Therefore, to create something, `annex` can be used. ~To inspect something `axiom` is used (and definitions can not be inspected by their `annex` identifier)?~

Annex is just a name of the form `%a.b.c`. An annex name can now be used for:
* `.ax`
* `.let`
* `.lam`/`.con`/`.fun`
* `.Pi`
* `.Sigma`
You can use any decl using an annex from C++ via:
* `world.call`: a curried call including type inference
* `world.annex`: retrieves the entitiy and that's it

`match`/`force` is still only supported for Axioms. We could add `match/`force/ for any annex, if we wanted to but this is a bit questionable but not completely idiotic either as I pointed out in Discord.

thorin/world.h
leissa: I originally intended to use the term *internal* instead of *annex*. That was the reason why I removed `make_internal`. But then, we decided that *internal* is a confusing name and went with *annex*. Is this OK?
```c++
make_external(bool on = true) { /* ... */ }
make internal() { make_external(false); }
```

NeuralCoder3: I think the name `make_external` implies that the def becomes external.
Maybe `make_external` and `make_internal` could take no bool and just call a function (named for instance `set_external[ity]`) with a bool.

dialects/core/be/ll.cpp
```diff
@@ -898,16 +906,10 @@ std::string Emitter::emit_bb(BB& bb, const Def* def) {
...
-        switch (exp.id()) {
-            case math::exp::exp:  f += "exp" ; break;
-            case math::exp::exp2: f += "exp2"; break;
-            case math::exp::log:  f += "log" ; break;
-            case math::exp::log2: f += "log2"; break;
-            case math::exp::log10: f += "log10"; break;
-        }
-        // clang-format on
+        f += (exp.sub() & sub_t(math::exp::Lbb)) ? "log" : "exp";
+        f += (exp.sub() & sub_t(math::exp::lbB)) ? "2" : (exp.sub() & sub_t(math::exp::lBb)) ? "10" : "";
```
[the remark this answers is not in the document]
leissa: I simply used more aliases. Is this fine?

thorin/world.h
leissa: I just went with `make_external()`/`make_insternal`.

PR #214, opened 2023-05-08
Added support for log10, abs and rounding

betzinger:


dialects/math/math.thorin
```diff
@@ -121,14 +121,15 @@
...
-.ax %math.exp(bl = exp, bL = log, Bl = exp2 = b, BL = log2):
+/// | Subtag | Alias       | B | L | T | Meaning                 | Semantics       |
+/// |--------|-------------|---|---|---|-------------------------|-----------------|
+/// | `bl`   | `exp`       | o | o | o | natural exponential     | \f$e^x\f$       |
+/// | `bL`   | `log`       | o | x | o | natural logarithm       | \f$\ln x\f$     |
+/// | `Bl`   | `exp2`, `b` | x | o | o | exponential with base 2 | \f$2^x\f$       |
+/// | `BL`   | `log2`      | x | x | o | logarithm with base 2   | \f$\log_2 x\f$  |
+/// | `BLT`  | `log10`     | x | o | x | logarithm with base 10   | \f$log_{10} x\f$|
+///
+.ax %math.exp(bl = exp, bL = log, Bl = exp2 = b, BL = log2, BLT=log10):
```
NeuralCoder3, line 132: The `T` parameter deems a bit out of place/artificial.
Maybe the option for the first letter `NBD` for `natural`, `binary`, and `decimal` would be a better fit.

dialects/math/math.thorin
```diff
@@ -153,6 +154,25 @@
 ///
 .ax %math.gamma(t, l): Π.[pe: «2; .Nat»][.Nat][%math.F pe] -> %math.F pe, normalize_gamma;
 ///
+/// ### %math.abs
+///
+/// [Absolute value] (https://en.wikipedia.org/wiki/Absolute_value) Returns absolute value of the provided value
```
NeuralCoder3, line 159: The markdown would render as:
"Absolute value Returns the absolute value of the provided value."
https://anydsl.github.io/thorin2/math.html#autotoc_md132

Maybe
`Returns [absolute value](...) of the provided value.`
or
`[Absolute value](...) of a an floating point` 
would fit better

(Also, I am not sure if markdown can handle the space between the text and the hyperlink.)

dialects/math/math.thorin
```diff
@@ -153,6 +154,25 @@
...
+///
+.ax %math.abs: Π.[pe: «2; .Nat»][.Nat][%math.F pe] -> %math.F pe, normalize_abs;
+///
+/// ### %math.round
+///
+/// [Rounding down], [Rounding up], [Rounding to next integer] and [Rounding towards zero/Truncating] (https://en.wikipedia.org/wiki/Rounding)
```
NeuralCoder3, line 165: `[Text](Link)` is the markdown tag for a link.
Example: `[google](https://google.com)` becomes [google](https://google.com)

This comment probably does not render as expected.

Suggestion:
``` 
Common rounding operations on floating points:
- [Rounding down](https://en.wikipedia.org/wiki/Rounding#Rounding_down) (`f`loor)
- ...
````

dialects/math/math.thorin
betzinger: Yeah, I've already the errors in the markdown. Will fix them

dialects/math/math.thorin
betzinger: I was contemplating something like 
math.exp.l = exp
math.exp.L = natural log
math.exp.lt = exp2
math.exp.Lt = log2
math.exp.lT = *unused*
math.exp.LT = log10
However, I thought it would be a problem since it wouldn't be compatible with existing code that uses math.exp

betzinger:
Thanks for the feedback! 
Building the docs should work now. 
For the math.exp part I followed @NeuralCoder3 's suggestion and set the base as first letter of the subaxioms with the options being N for natural, B for binary and D for decimal

PR #216, opened 2023-05-12
Bugfix/ds2csp

leissa:
fixes bug in ds2cps + test case

#210, @fodinabor: may also be the issue you encountered


fodinabor (approved):
Yes, does seem to fix my imminent problem (still not magically running through now, but I assume that means I have to fix stuff on my side :))

PR #217, opened 2023-05-13
type check lambdas when last op (body) is set

leissa:
* finally type checking for mutable Lams
* fixed bug in alpha-equiv - related to #174?
* fixed bug in lower_matrix_mediumlevel.cpp (wrong filter type)
* fixed type inference for things like `umax(0, ?) == 1`

### DIFF
--- a/dialects/matrix/passes/lower_matrix_mediumlevel.cpp
+++ b/dialects/matrix/passes/lower_matrix_mediumlevel.cpp
@@ -250,7 +250,7 @@ Ref LowerMatrixMediumLevel::rewrite_(Ref def) {
             iterator[idx]               = world.call<core::bitcast>(world.type_idx(dim_nat_def), iter);
             auto [new_mem, new_mat]     = new_acc->projs<2>();
             acc                         = {new_mem, new_mat};
-            current_mut->set(false, for_call); // TODO correct filter?
+            current_mut->set(true, for_call);
             current_mut = body;
         }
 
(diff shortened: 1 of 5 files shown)


leissa:
@NeuralCoder3 can you double-check my changes in `lower_matrix_mediumlevel.cpp`?

dialects/matrix/passes/lower_matrix_mediumlevel.cpp
```diff
@@ -250,7 +250,7 @@ Ref LowerMatrixMediumLevel::rewrite_(Ref def) {
             iterator[idx]               = world.call<core::bitcast>(world.type_idx(dim_nat_def), iter);
             auto [new_mem, new_mat]     = new_acc->projs<2>();
             acc                         = {new_mem, new_mat};
-            current_mut->set(dim_nat_def, for_call);
+            current_mut->set(false, for_call); // TODO correct filter?
```
NeuralCoder3, line 253: The `dim_nat_ def?` filter was wrong. 
`false` is okay. But `true` as a filter should also work.
These functions just call the for-loop (they are just wrappers).
Therefore, we could always unfold the wrapper.

NeuralCoder3:
The build of [bfdcd46](https://github.com/AnyDSL/thorin2/pull/217/commits/bfdcd46e5bd96ca3ead6140e13572b733587e32c) seems to fail. Maybe the filter can not be `true`?

PR #218, opened 2023-05-17
Refactor/colon

leissa:
Use `:` in `.lam`/`.lm`/`.fun`/`.fn` to separate (optional) return type instead of `->`:
```
.lam f(x y: .Nat): .Nat = x;
```
This is like in Lean or SML and resolves an ambiguity:
```
.lam f x: .Nat -> .Nat = x; 
// does it mean: 
.lam f(x: .Nat) -> .Nat = x;
// or:
.lam f(x: .Nat -> .Nat) = x;
```
And looks like this now:
```
.lam f x: .Nat : .Nat = x;
```
You still want to use parentheses in this case though...

### DIFF
--- a/dialects/matrix/matrix.thorin
+++ b/dialects/matrix/matrix.thorin
@@ -146,7 +146,7 @@
 ///
 /// Follow the principle `ij <- ik,kj` (`out[i,j] = sum_k in1[i,k] * in2[k,j]`) by using mulplication as combination function and addition as reduction function.
 .lam .extern internal_mapRed_matrix_prod
-    ![m: .Nat, k: .Nat, l: .Nat, [p: .Nat, e:.Nat]] ->
+    ![m: .Nat, k: .Nat, l: .Nat, [p: .Nat, e:.Nat]]:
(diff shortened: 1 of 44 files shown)


dialects/matrix/matrix.thorin
```diff
@@ -146,7 +146,7 @@
 ///
 /// Follow the principle `ij <- ik,kj` (`out[i,j] = sum_k in1[i,k] * in2[k,j]`) by using mulplication as combination function and addition as reduction function.
 .lam .extern internal_mapRed_matrix_prod
-    ![m: .Nat, k: .Nat, l: .Nat, [p: .Nat, e:.Nat]] ->
+    ![m: .Nat, k: .Nat, l: .Nat, [p: .Nat, e:.Nat]]:
```
NeuralCoder3, line 149: In general, it seems more readable / I am more used to `: return`.
However, in this case, it seems a bit odd (well the large return type itself is probably odd).

I think it is an improvement overall.

NeuralCoder3 (approved):

PR #219, opened 2023-05-18
Alpha equiv 2 - Electric Boogaloo

### LINKED ISSUE #109: Performance depends on the size of Arr/pack

leissa:
Alpha-Equivalence checks now knows two modes:
* In `infer` mode, type inference is happening and Infer%s will be resolved, if possible.
    Also, two *free* but *different* Var%s **are** considered α-equivalent.
* When **not* in `infer` mode, no type inference is happening and Infer%s will not be touched.
    Also, Two *free* but *different* Var%s are **not** considered α-equivalent.

In addition:
* tuples/sigma are now converted to packs/arrs modulo α-equivalence:
    * `(a, a')` => `‹2; a›`
    * `[A, A']` => `«2; A»`
* fixes #109
* should also fix the related issue in #210


thorin/check.h
```diff
@@ -38,41 +38,33 @@ class Infer : public Def {
...
-    World& world() const { return *world_; }
-
-    /// Are @p d1 and @p d2 α-equivalent?
-    bool equiv(Ref d1, Ref d2);
+    /// Are d1 and d2 α-equivalent?
+    /// * In @p infer mode, type inference is happening and Infer%s will be resolved, if possible.
+    ///     Also, two *free* but *different* Var%s **are** considered α-equivalent.
+    /// * When **not* in @p infer mode, no type inference is happening and Infer%s will not be touched.
+    ///     Also, Two *free* but *different* Var%s are **not** considered α-equivalent.
+    template<bool infer = true> static bool alpha(Ref d1, Ref d2) { return Check().alpha_<infer>(d1, d2); }
```
fodinabor, line 51: So we loose basically all caching across two `alpha` calls?

thorin/check.cpp
```diff
@@ -140,70 +140,61 @@ bool Checker::equiv_internal(Ref d1, Ref d2) {
...
-}
-
-/*
- * infer
- */
-
-const Def* Pi::infer(const Def* dom, const Def* codom) {
-    auto& w = dom->world();
-    return w.umax<Sort::Kind>({dom->unfold_type(), codom->unfold_type()});
+    // outln("{}", first);
```
fodinabor, line 184: cleanup? :)

fodinabor (approved):
Jup, this correctly types the `(f, (ff, t)#done)#matched arg`  thing, so one can call it.

(Backend still doesn't like these things ^^)

NeuralCoder3 (approved):

thorin/check.h
leissa: yes, but I always found the caching questionable to begin with as we could mutate the alpha-equivalence away. We can later on still try to add some caching mechanism if this is too slow.

PR #221, opened 2023-05-22
removed doxygen workaround

leissa:
supersedes #183


leissa:
Unfortunately, we are facing a new Doxygen problem ...

PR #225, opened 2023-05-24
implemented proper parsing of packs

leissa:
+ test case
+ small bug fix in cps2ds

### DIFF
--- a/thorin/fe/parser.cpp
+++ b/thorin/fe/parser.cpp
@@ -312,24 +312,34 @@ Ref Parser::parse_arr_expr() {
 }
 
 Ref Parser::parse_pack_expr() {
-    // TODO This doesn't work. Rework this!
     auto track = tracker();
     scopes_.push();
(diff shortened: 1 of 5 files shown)


thorin/fe/parser.cpp
```diff
@@ -312,24 +312,34 @@ Ref Parser::parse_arr_expr() {
...
     eat(Tag::D_angle_l);
 
-    const Def* shape;
-    // bool mut = false;
     if (ahead(0).isa(Tag::M_id) && ahead(1).isa(Tag::T_colon)) {
         auto id = eat(Tag::M_id);
         eat(Tag::T_colon);
 
-        shape      = parse_expr("shape of a pack");
+        auto shape = parse_expr("shape of a pack");
```
NeuralCoder3, line 323: Just move this line in front of the if (it occurs in the if and afterward).

NeuralCoder3 (approved):

PR #229, opened 2023-05-31
new partial eval strategy

leissa:
* partial eval is now performed on the fly during construction of an App
* solves beta-equiv problem
* test case
* removes old PartialEval RWPass

@NeuralCoder3: I changed one filter in `reshape.cpp` from `.ff` to `.tt`. Can you double-check whether this is a show stopper or so?


NeuralCoder3:
> @NeuralCoder3: I changed one filter in `reshape.cpp` from `.ff` to `.tt`. Can you double-check whether this is a show stopper or so?

It should be okay. It should also be possible to copy the filter from the original function (maybe taking care of dependencies).

The true filter probably was a relict from when we wrapped every function for reshape.

docs/langref.md
```diff
@@ -337,20 +337,21 @@ The following table summarizes the different tokens used for functions declarati
 
 The following function *declarations* are all equivalent:
```
NeuralCoder3, line 338: We could maybe mention explicitly in the documentation that the default filter for lams is true.

leissa: yeah, the docs only barely mention the filter at all. will write more on that, when I have more time

> > @NeuralCoder3: I changed one filter in `reshape.cpp` from `.ff` to `.tt`. Can you double-check whether this is a show stopper or so?
> 
> It should be okay. It should also be possible to copy the filter from the original function (maybe taking care of dependencies).
> 
> The true filter probably was a relict from when we wrapped every function for reshape.

ah, I've now done it the correct way by rewriting the old one.

PR #232, opened 2023-06-07
type-check let

leissa:
Implemented missing type check for let decl

### DIFF
--- a/thorin/fe/parser.cpp
+++ b/thorin/fe/parser.cpp
@@ -882,25 +882,34 @@ void Parser::parse_ax_decl() {
 }
 
 void Parser::parse_let_decl() {
+    auto track = tracker();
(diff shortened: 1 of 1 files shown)


fodinabor:
So this should allow `.let [a:%core.i32, b:%core.i8] = (48:%core.i32, 3:%core.i8);` and would complain about `.let [a:%core.i32, b:%core.i8] = (48:%core.i32, 3:%core.i32);` ?

leissa:
> So this should allow `.let [a:%core.i32, b:%core.i8] = (48:%core.i32, 3:%core.i8);` and would complain about `.let [a:%core.i32, b:%core.i8] = (48:%core.i32, 3:%core.i32);` ?

yes

PR #237, opened 2023-06-21
Fixes

### LINKED ISSUE #236: Parametrized packs fail for arity <= 1
If you try to create a parametrized pack `‹i: n, i›` with arity `n <= 1` as shown in the example below, Thorin will run into the following assertion:
```
thorin: thorin/util/cast.h:23: T* thorin::RuntimeCast<B>::as() [with T = thorin::Arr; B = thorin::Def]: Assertion `isa<T>()' failed.
Aborted
```
#### Example
```rust
.let p1 = ‹i: 2; i›; // ok: (0, 1)
.let p2 = ‹i: 1; i›; // fail, should be: (0,) = 0
.let p3 = ‹i: 0; i›; // fail, should be: ()
```

leissa:
Various fixes and improvements
* fixed a weird edge case when checking alpha-equiv for mutables with same arity ...
* fixed #236
* new CLI switch `--break-on-alpha-unequal` which helps a lot to debug probls 
* some cleanup


thorin/check.h
```diff
@@ -23,13 +23,15 @@ class Infer : public Def {
     Infer* unset() { return Def::unset()->as<Infer>(); }
     ///@}
 
-    /// @name union-find
-    ///@{
+    /// Eliminate Infer%s that have may have been resolved in the meantime by rebuilding.
```
NeuralCoder3, line 26: "Eliminate Infer%s that ~~have~~ may have been resolved in the meantime by rebuilding"

thorin/check.cpp
```diff
@@ -42,16 +59,35 @@ const Def* Infer::find(const Def* def) {
...
+        }
+        return true;
+    }
+    return false;
+}
 /*
  * Check
  */
 
+#ifdef THORIN_ENABLE_CHECKS
```
NeuralCoder3, line 78: Isn't there missing a case to define `bool Check::fail` if `THORIN_ENABLE_CHECKS` isn't defined?

NeuralCoder3 (approved):

thorin/check.cpp
leissa: This is inline in the header

PR #241, opened 2023-07-04
introduced threshold for Scalerize Pass

leissa:
If you are juggling around large tuples, compile time will explode. This threshold, will not scalarize tuples larger than the given threshold.


fodinabor (approved):
Damn, I like the use of a frontend constant for this ^^

leissa:
> Damn, I like the use of a frontend constant for this ^^

I thought about a `%cli` dialect that allows you to define variables. Sth like this:
```sh
thorin foo.thorin -D "scalerize_threshold = 32"
```

PR #243, opened 2023-07-06
Refactor/regex

leissa:
This massages the regex code a bit.
* `regex.thorin`
    * use %core.idx instead of casts
    * use %core.mode.nuw for %core.wrap.add and %core.idx
      I actually experienced some serious performance issues some time ago, if this was missing, but I checked with the benchmark: doesn't do anything in our case
    * use more type aliases/functions
    * make use of ptrn groups
    * removed superfluous braces: `{`, `}`
    * use escaped chars (`\t` etc) instead of magic numbers
    * `1:.Bool` -> `.tt`
* C++ files
    * use `world.call` instead of `world.annex` + `world.app`
    * removed now superfluous `any_impl` & co

### DIFF
--- a/dialects/regex/passes/lower_regex.cpp
+++ b/dialects/regex/passes/lower_regex.cpp
@@ -15,21 +15,6 @@ Ref rewrite_arg(Ref ref, Ref n);
 
 Ref wrap_in_cps2ds(Ref callee) { return direct::op_cps2ds_dep(callee); }
 
-Ref any_impl(Match<regex::any, Axiom> any_ax) {
-    auto& world = any_ax->world();
-    return world.annex<regex::match_any>();
-}
-
-Ref range_impl(Match<regex::range, App> range_app) {
-    auto& world = range_app->world();
-    return world.call<regex::match_range>(range_app->arg());
-}
-
-Ref not_impl(Match<regex::not_, App> not_app) {
-    auto& world = not_app->world();
-    return world.annex<regex::match_not>();
-}
-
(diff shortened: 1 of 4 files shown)


dialects/regex/passes/lower_regex.cpp
```diff
@@ -66,16 +31,16 @@ Ref rewrite_arg(Ref def, Ref n) {
...
-    if (auto not_app = thorin::match<not_>(def))
-        new_app = world.iapp(world.app(not_impl(not_app), n), rewrite_args(not_app->arg(), n));
-    if (auto conj_app = thorin::match<conj>(def))
-        new_app = world.iapp(world.app(conj_impl(conj_app), n), rewrite_args(conj_app->arg(), n));
-    if (auto disj_app = thorin::match<disj>(def))
-        new_app = world.iapp(world.app(disj_impl(disj_app), n), rewrite_args(disj_app->arg(), n));
-    if (auto quant_app = thorin::match<quant>(def))
-        new_app = world.iapp(world.app(quant_impl(quant_app), n), rewrite_args(quant_app->arg(), n));
+    // clang-format off
+    if (auto _any   = thorin::match<any  >(def)) new_app = world.call<regex::match_any  >(n);
```
fodinabor, line 35: don't start names with `_`... that's reserved.

fodinabor (commented):
Overall looking good, just a nit.

leissa:
brings us down from 323 -> 280 C++ loc :)

fodinabor:
G'old LOC tuning :P

PR #245, opened 2023-07-19
Simpler loop from %affine.For

### LINKED ISSUE #244: %affine.For breaks without scalerize pass
#### Example: `bug.thorin`
The Thorin function `foo` computes the sequence 0, 1, ..., 99 as a pack `«100; %core.I32»` via a simple `%affine.For` counting loop. Initially, the pack `‹100; 0:%core.I32›` is passed to this loop and updated as an accumulator `acc` between iterations. Inside the function `bar`, said pack is finally stored at some `ptr: %mem.Ptr0`.

```rust
.plugin core;
.plugin affine;

.let Arr = «100; %core.I32»;

.fun foo []: Arr =
    .con body [i: %core.I32, acc: Arr, yield: .Cn Arr] =
        .let `acc = .ins (acc, %core.conv.u 100 i, i);
        yield acc;

    .let loop = %affine.For (%core.i32, 1, Arr);
    loop (0:%core.I32, 100:%core.I32, 1:%core.I32, ‹100; 0:%core.I32›, body, return);

.fun .extern bar [mem: %mem.M, ptr: %mem.Ptr0 Arr]: %mem.M =
    .ret arr  = foo $ ();
    .let `mem = %mem.store (mem, ptr, arr);
    return mem;
```

#### Example: `bug.c`
In addition to that, the following C program calls `bar` with a pointer to an array and prints its contents to `stdout`.

```c
#include <stdio.h>
#include <stdint.h>

extern void bar(uint32_t* arr);

int main() {
    uint32_t arr[100] = {0};
    bar(arr);

    for (size_t i = 0; i < 100; i++)
        printf("%i ", arr[i]);
    printf("\n");

    return 0;
}
```

#### Commands
```
thorin bug.thorin --output-ll bug.ll
clang -o test bug.c bug.ll -Wno-override-module
./test
```

However, this only produces the correct output when the scalerize pass is run which flattens the pack inside the loop. In case it is skipped (e.g. `%compile.scalerize_threshold` is exceeded), the output contains garbage instead.

Looking at the generated LLVM IR code, it seems `acc` is not updated correctly when jumping between BBs that form the loop.

leissa:
This patch creates a much simpler and more straightforward loop in `LowerFor` for `%affine.For`. Previously, the loop was quite dependent on certain optimizations to happen. Otherwise, some funny things can end up in the backend.  

*Remark:* We probably want to reduce this pass even further to just substitute `%affine.For` with an appropriate loop as I sketched in the comment. But rn, we can't quite do it, if the accumulator contains a `mem`. 

* Fixes #244


dialects/affine/passes/lower_for.h
```diff
@@ -6,6 +6,25 @@
...
+///         .con new_body() = body (iter, acc, .cn acc: «i: n; Ts#i» =
+///             .let `iter = %core.wrap.add %core.mode.nsuw (iter, step);
+///             head (iter, acc));
+///         .con new_exit() = exit (acc);
+///         (new_exit, new_body)#(%core.icmp.ul (iter, end)) ();
+///     head(begin, init);
+/// ```
+/// However, we merge `init`/`acc` into the signature, as it may contain a `mem`.
+/// In this case we have to equip `new_body`/`new_exit` with a `mem` as well.
+/// @todo We probably want to have phases which fix such things so we don't have to do this in C++.
```
NeuralCoder3, line 27: If I understand the problem correctly, the memory passes like reshape and add_mem should be these passes.

NeuralCoder3 (approved):

dialects/affine/passes/lower_for.h
leissa: I discovered another problem regarding performance. If we have `acc: «1000; T»`, we don't want to merge this with `iter` ...
I'm currently working on things like that, but I'm doing this in a different branch in another PR.

leissa: The problem are the branches:
```
.con foo(mem: %mem.M, ...) =
    .con t() = load (mem, ...)
    .con f() = store (mem, ...)
    (f, t)#cond ()
```
This is kind of dangerous, because `load`/`store` could now float out of `t`/`f` causing all kind of weird problems. For this reason, my patch in this PR looks for a `mem` and puts it into the `head`er's branch argument. Does AddMem/Reshape cope with that?

PR #251, opened 2023-10-02
Refactor/fe

leissa:
* This PR factors out various parts of thorin to use [FE](https://leissa.github.io/fe/) instead. 
* It also includes #250.
* `modules` -> `external` to obey [pitchfork](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs) directory conventions


leissa:
@fodinabor can you have a look at the failing macos test cases? I upgraded to `macos-latest` I your `FileCheck` magic needs an update.

leissa:
@fodinabor , @NeuralCoder3 : anyone has the mercy to give their 2 cents on this? :)

thorin/fe/lexer.cpp
```diff
@@ -39,16 +39,15 @@ Tok Lexer::lex() {
             return *cache;
         }
 
-        loc_.begin = ahead().pos;
-        str_.clear();
+        begin();
```
fodinabor, line 42: `begin` might not be the best name, as it's quite heavily associated with iterators in C++. maybe prefer `start`?

thorin/fe/lexer.cpp
```diff
@@ -39,16 +39,15 @@ Tok Lexer::lex() {
...
 
 #if defined(_WIN32) && !defined(NDEBUG) // isspace asserts otherwise
         if (accept_if([](int c) { return (c & ~0xFF) == 0 ? isspace(c) : false; })) continue;
 #else
         if (accept_if(isspace)) continue;
 #endif
-        if (accept(utf8::Err)) error(loc_, "invalid UTF-8 character");
-        if (accept(utf8::EoF)) return tok(Tag::M_eof);
+        if (accept(0)) error(loc_, "invalid UTF-8 character");
```
fodinabor, line 49: why introduce a new magic constant when you had an enum for it?
if you mean `\0`, maybe use that?

thorin/util/dl.h
```diff
@@ -7,8 +7,8 @@ namespace thorin::dl {
 
 std::string_view extension(); ///< `".dll"` or `".so"`
 
-void* open(const std::string& filename);
-void* get(void* handle, const std::string& symbol_name);
+void* open(const char* filename);
```
fodinabor, line 10: can we agree on `string_view`?

fodinabor (commented):
From what I can see, it looks very NFC, which is probably what you want, so, LGTM.
Just a few nits.

thorin/util/dl.h
leissa: The problem is that the APIs underneath that expect a `const char*`. So, changing to `std::string_view` means, that we would have to *convert* the original string - which **is** already a `const char*` to `std::string_view` just to *convert* this thing back to `const char*` - which is btw a bit annoying to do:
```c++
void* open(std::string_view sv) {
    std::string str(sv);
    auto file = str.c_str();
    // ...
}
```
This all seemed a bit idiotic to me. Hence, I just used `const char*`. But I do see the consistency argument. So, if you insist, I'll change it to `std::string_view` :)

fodinabor: Aah.. Right. That's been the issue, why we're not using string view here ^^ its OK :)

PR #254, opened 2023-10-30
Implement codegen from regex DFA.

fodinabor:
For this, LowerRegex first creates a NFA from the Regex, then converts this to an minimized DFA.
Finally, this DFA is emitted as Thorin code again (see dfa2matcher).

This allows us to match non-deterministic regex such as `(a|b)+a` correctly as well, which was a pretty strong downside of the old impl.
Note, we still don't support capture groups and the like, and also, we lose support for `not` although that might be reimplementable..
Also, we lost the showcase of using the annexes as implementation.

The regex->min dfa infra follows the algorithms from https://doi.org/10.1007/978-3-642-17540-4 and hopcroft from https://en.wikipedia.org/wiki/DFA_minimization.

A super quick&dirty benchmark seems favorable :)
(note how the old impl was slower than manual and the new one is faster.. :))

old:
|engine|average[us]|min[us]|max[us]|deviation[%]|
|-|-|-|-|-|
|thorin|1069|1055|1096|3|
|ctre|4640|4606|4696|1|
|pcre2|3723|3603|3870|7|
|manual|843|821|878|6|
|pcre2_jit|1233|1210|1269|4|
|std|9692|9607|9819|2|

new:
|engine|average[us]|min[us]|max[us]|deviation[%]|
|-|-|-|-|-|
|ctre|4663|4656|4709|1|
|pcre2|3769|3645|3848|5|
|manual|865|849|904|6|
|thorin|738|725|762|5|
|std|10227|10171|11174|9|
|pcre2_jit|1224|1211|1258|3|


dialects/regex/pass/dfa.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+using namespace thorin::automaton;
+
+void DFANode::add_transition(const DFANode* to, std::uint16_t c) {
+    if (auto it = transitions_.find(c); it != transitions_.end())
+        it->second = to;
+    else
+        transitions_.emplace(c, to);
+}
+
```
leissa, line 15: why not simply `transitions_[c] = to`?

dialects/regex/pass/dfa.h
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    bool is_accepting() const { return accepting_; }
+    void set_accepting(bool accepting) { accepting_ = accepting; }
+
+    void dump() const;
+
+private:
+    std::unordered_map<std::uint16_t, const DFANode*> transitions_;
```
leissa, line 40: `absl::flat_hash_map<std::uint16_t, const DFANode*>`

dialects/regex/pass/dfa2matcher.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+// the main idea is:
+// for every state in the DFA, we create a lam that checks if the char c at pos is the end of the string \0
+// if not, jump to a checker lam, that consists of a few bit-wise or-ed in-range checks to verify if we can transition
+// to a certain other state with c if any of the checks is true, we jump to the corresponding state lam with the updated
+// position if the check fails, we jump to the next checker lam or the exit lam if we checked all possible transitions
+// Note, since the DFA stores the transitions as chars, not ranges, we have to merge the transitions to ranges first
+// (cf. transitions_to_ranges)
+
+#define DEBUG_PRINT 0
```
leissa, line 28: use `world.DLOG(...)` instead

dialects/regex/pass/nfa2dfa.cpp
```diff
@@ -0,0 +1,82 @@
...
+            if (nfaState->is_accepting()) {
+                dfaState->set_accepting(true);
+                break;
+            }
+        }
+    }
+    return dfa;
+}
+
+} // namespace thorin::automaton
```
leissa, line 82: newline

dialects/regex/range_helper.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include <optional>
+#include <vector>
+
+#include "thorin/world.h"
+
+namespace thorin::regex {
+std::vector<std::pair<nat_t, nat_t>> merge_ranges(World& w, const std::vector<std::pair<nat_t, nat_t>>& old_ranges);
+
+} // namespace thorin::regex
```
leissa, line 10: newline

lit/regex/match_manual.thorin
```diff
@@ -0,0 +1,84 @@
...
+            .let is_end = %core.icmp.e (c, 0:I8);
+            (not_end, accept)#is_end (mem, new_i);
+
+        start(mem, 0:I32)
+    };
+
+    (exit, match_argument) # (%core.icmp.ug (argc, 1:I32)) (mem, 0:I32)
+};
+
+// CHECK-NOT: %regex.
```
leissa, line 84: newline

leissa (commented):
Fantastic work :)
I haven't read it line by line, but looks pretty good to me.

dialects/regex/pass/dfa2matcher.cpp
fodinabor: Hmm... I'm a bit hesitant to "clutter" the pure automaton impl. stuff with thorin implementation stuff, i.e. passing through the world just to be able to debug print feels a bit off.. but dunno...

leissa: I just grep'ed through the code. There are three instances of `#if DEBUG_PRINT` - each of those has the `World` available anyway...

PR #257, opened 2023-11-08
Also support regex.not_ in DFA-based engine.

fodinabor:
This introduces the notion of error states to the DFA/NFA that trap negated matches.


leissa:
merge?

PR #260, opened 2023-11-22
Refactor/absl array

leissa:
This PR removes our home-brew `Array<T>` and `Span<T>` classes with `absl::InlinedVector<T>` and a new `Span<T>` which is only a thin wrapper for `std::span<T>`.

# Span<T, N>

This is a thin wrapper for `std::span<T, N>` with the following additional features:
* Constructor with `std::initializer_list` (C++26 will get this ...)
* Constructor for any compatible Vectorlike argument
* rsubspan (reverse subspan)
* structured binding, if `N != std::dynamic_extent:`
    ```c++
    void f(Span<int, 3> span) {
        auto& [a, b, c] = span;
        b = 23;
        // ...
    }
    ```
    This will save us a few copies as we formerly copied over to `std::array` in order to be able to decompose the array.

# Vector<T>

This thing behaves like a `std::vector` but is under the hood an `absl::InlinedVector`. The number of *inlined* elements is 4 for most element types.

# Misc

* No more `DefArray` vs `DefVec`. Just use `DefVec` = `Vector<const Def*>`
* No more confusion with `Array<T>` (the C++ data structure) vs `Arr` (the thorin type) - `Array<T>` is gone
* There is subtle difference with our old `Span` class vs our new one: the old one always used `const T` internally. Now, you have to do either `Span<const T>` or `View<T>` (an alias).
* The lambda generator constructor is now a free function:
    ```c++
    auto vec = vector<int>(23 ,[](size_t i) { return i; });
    ```
* I replaced a few places with explicit loops (`rewrite.cpp`) as stepping through `std::function` and friends during debugging becomes super annoying.


fodinabor (commented):
LGTM.
Only the nit that I find the `vector<..>()` helper a bit confusing as I am then looking for `using namespace std` or `using std::vector` to identify if this is coming from `std::vector` - maybe `make_Vector`? :)

leissa:
> LGTM. Only the nit that I find the `vector<..>()` helper a bit confusing as I am then looking for `using namespace std` or `using std::vector` to identify if this is coming from `std::vector` - maybe `make_Vector`? :)

Alright, I now subclassed `absl::InlinedVector` instead of aliasing it. So the original c'tors are back via:
```C++
auto vec = Vector<int>(23, [](size_t i) { return i; });
```
There btw seems to be a gcc problem for this:
```c++
auto vec = Vector<int>(my_range_of_ints, [](auto i) { return i; });
```
It's a subtle issue with template instantiation, overloading, concepts, etc. I'm pretty sure it's a gcc bug. Works with clang. As a workaround, you have to explicitly specify the argument:
```c++
auto vec = Vector<int>(my_range_of_ints, [](int i) { return i*i; });
```


Subclassing gives us another benefit to have `vec.span()` and `vec.view()` methods, which are a bit nicer to use:
```c++
for (auto elem : vec.view().subspan(1)) { /*...*/ }
```
instead of
```c++
for (auto elem : View(vec).subspan(1)) { /*...*/ }
```

PR #269, opened 2024-03-17
Vars

leissa:
This is sth I wanted to do for years, but I never really figured out how to do that exactly: 
I want to get rid of `thorin::Scope` and, instead, work with free Vars. This is a major milestone with the new subsystem working but the old stuff is still in place - although I've already ported some of it. So, expect more PRs to come.

* `Def::local_muts()`/`Def::local_vars()` are `Var`s/mutables reachable by following *immutable* `Def::extended_ops()`.
* `Def::local_muts()`/`Def::local_vars()` are cached and hash-consed.
* `Def::free_vars()` compute a global solution, i.e., by transitively following *mutables* as well.
* `Def::free_vars()` are computed on demand and cached.
 They will be transitively invalidated by following `fv_consumers()`, if a mutable is mutated.
* `Def::is_closed()`/`Def::is_open()` to check whether a `Def` contains free vars.
 
 ## Change log

* removed `ScopeRewriter` in favor of new `free_var`-based `VarRewriter`
* simplifed `rewrite` interface
* `PooledSet` and `Pool` to hash-cons and arena-allocate sets
* sanity check whether externals are closed
* fixed issue in `cps2ds.cpp` but one issue remains (#271) 
* minor bug fixes/additions here & there

## Future directions

* lazy rewrite/substitution (only substitute one mutable and a time and continue substituting as needed to prevent quadratic runtime when resolving curried beta-reduction etc)
* completely remove `thorin::Scope`
* new textual outputs
    * a simple output that dumps everything up to mutables
    * a proper "decompilation" of the thorin Graph into a thorin AST


leissa:
@fodinabor, @NeuralCoder3 bump

lit/direct/ad_mem.thorin.disabled2
NeuralCoder3, line 1: Does this test no longer work?

include/thorin/util/pool.h
NeuralCoder3, line 1: If I see it correctly, this code is completely independent of Thorin, right?
Is the thorin namespace the correct place for it?

src/thorin/def.cpp
```diff
@@ -77,36 +90,37 @@ UMax::UMax(World& world, Defs ops)
...
-Ref Sigma    ::rebuild(World& w, Ref  , Defs o) const { return w.sigma(o)                     ->set(dbg()); }
-Ref Singleton::rebuild(World& w, Ref  , Defs o) const { return w.singleton(o[0])              ->set(dbg()); }
-Ref Type     ::rebuild(World& w, Ref  , Defs o) const { return w.type(o[0])                   ->set(dbg()); }
-Ref Test     ::rebuild(World& w, Ref  , Defs o) const { return w.test(o[0], o[1], o[2], o[3]) ->set(dbg()); }
-Ref Tuple    ::rebuild(World& w, Ref t, Defs o) const { return w.tuple(t, o)                  ->set(dbg()); }
-Ref UInc     ::rebuild(World& w, Ref  , Defs o) const { return w.uinc(o[0], offset())         ->set(dbg()); }
-Ref UMax     ::rebuild(World& w, Ref  , Defs o) const { return w.umax(o)                      ->set(dbg()); }
-Ref Var      ::rebuild(World& w, Ref t, Defs o) const { return w.var(t, o[0]->as_mut())       ->set(dbg()); }
-Ref Vel      ::rebuild(World& w, Ref t, Defs o) const { return w.vel(t, o[0])->set(dbg())     ->set(dbg()); }
+Ref Idx      ::rebuild(World& w, Ref  , Defs  ) const { assert(isa_imm()); return w.type_idx(); }
```
NeuralCoder3, line 101: Can't we get rid of the redundancy here?
Or are the checks and calls too irregular or need to be at this source location for clarity?

src/thorin/fe/lexer.cpp
```diff
@@ -275,6 +275,7 @@ void Lexer::parse_digits(int base /*= 10*/) {
     }
 }
 
+// TODO do we need this?
```
NeuralCoder3, line 278: https://github.com/search?q=repo%3AAnyDSL/thorin2%20_any&type=code
(Although the search can be unreliable.)

I don't think so. (One would need to test what depends on the function)

src/thorin/pass/pass.cpp
```diff
@@ -39,7 +39,7 @@ void PassMan::pop_states(size_t undo) {
 }
 
 void PassMan::run() {
-    world().ILOG("run");
+    world().verify().ILOG("run");
```
NeuralCoder3, line 42: Are there cases when we don't want to verify before ILOG?

src/thorin/plug/direct/pass/cps2ds.cpp
```diff
@@ -129,12 +129,25 @@ const Def* CPS2DS::rewrite_body_(const Def* def) {
...
     if (def->isa<Var>()) return def;
+    if (def->isa<Global>()) return def;
 
     if (auto tuple = def->isa<Tuple>()) {
         auto elements = DefVec(tuple->ops(), [&](const Def* op) { return rewrite_body(op); });
         return world().tuple(def->type(), elements)->set(tuple->dbg());
     }
 
+    // TODO there are more probls like this
+    if (auto old_mut = def->isa_mut()) {
```
NeuralCoder3, line 140: Is there a high-level view of what is going on here/what the problem is
for the next one reading this code?

NeuralCoder3 (approved):

lit/direct/ad_mem.thorin.disabled2
NeuralCoder3: This answers my question probably:
https://github.com/AnyDSL/thorin2/issues/271

include/thorin/util/pool.h
leissa: There is a `TODO` at the top of the file that I want to move it to `fe`. There are also some instances of code duplication with `fe::Sym` which aren't exactly easy to resolve because `fe::Sym` has a short string optimization and all the templates etc. For now, I think we can leave it like this.

lit/direct/ad_mem.thorin.disabled2
leissa: Yes, see #271

src/thorin/pass/pass.cpp
leissa: There may be cases where you are in the middle of rebuilding sth where an external gets some free variables before you mark it as internal.

PR #274, opened 2024-04-29
Refactor/blockexpr

leissa:
Plz first give #273 your blessing.

* removes block expressions: `{ d* e }`.
  simply use `d* e` instead and if you really want to use delimiters use a tuple: `(d* e)`
* `abc::(a: A, b: B, c: C)` -> `(a: A, b: B, c: C)::abc`.
  This is simpler to parse and makes more sense as the identifier `abc` is only available *after* having processed the whole tuple pattern.

### DIFF
--- a/include/thorin/ast/ast.h
+++ b/include/thorin/ast/ast.h
@@ -248,7 +248,7 @@ private:
 /// `dbg::(ptrn_0, ..., ptrn_n-1)` or `dbg::[ptrn_0, ..., ptrn_n-1]`
 class TuplePtrn : public Ptrn {
 public:
-    TuplePtrn(Loc loc, bool rebind, Dbg dbg, Tok::Tag delim_l, Ptrs<Ptrn>&& ptrns)
+    TuplePtrn(Loc loc, Tok::Tag delim_l, Ptrs<Ptrn>&& ptrns, bool rebind, Dbg dbg)
         : Ptrn(loc, rebind, dbg)
         , delim_l_(delim_l)
         , ptrns_(std::move(ptrns)) {}
--- a/include/thorin/ast/ast.h
+++ b/include/thorin/ast/ast.h
@@ -192,6 +192,16 @@ private:
     bool rebind_;
 };
 
+class ErrorPtrn : public Ptrn {
+public:
+    ErrorPtrn(Loc loc)
+        : Ptrn(loc, false, Dbg()) {}
+
+    void bind(Scopes&, bool quiet = false) const override;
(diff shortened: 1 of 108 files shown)


NeuralCoder3:
> abc::(a: A, b: B, c: C) -> (a: A, b: B, c: C)::abc.

I think this syntax is confusing.
`:` is used as type annotation. Hence, we used `::` as `name :: destructor pattern`.
Which was okay as it conceptually was somewhat similar to the type pattern.
However, with both swapped, it becomes even less clear.
Furthermore, `::` is commonly used for lists or type hints in Haskell.

I think a syntax like `as` would be clearer.
`(a: A, b: B, c: C) as abc` means we have the destructured tuple `(a,b,c)` but also identify the whole expression as `abc`.

gtest/test.cpp
```diff
@@ -1,38 +1,37 @@
...
-    auto r = parser.scopes().find({Loc(), driver.sym("r")});
+    // auto c = parser.scopes().find({Loc(), driver.sym("c")});
+    // auto r = parser.scopes().find({Loc(), driver.sym("r")});
 
-    EXPECT_TRUE(r->is_term());
-    EXPECT_TRUE(!r->type()->is_term());
-    EXPECT_EQ(c, r);
+    // EXPECT_TRUE(r->is_term());
+    // EXPECT_TRUE(!r->type()->is_term());
+    // EXPECT_EQ(c, r);
```
NeuralCoder3, line 34: Are these lines no longer true?

leissa:
> > abc::(a: A, b: B, c: C) -> (a: A, b: B, c: C)::abc.
> 
> I think this syntax is confusing. `:` is used as type annotation. Hence, we used `::` as `name :: destructor pattern`. Which was okay as it conceptually was somewhat similar to the type pattern. However, with both swapped, it becomes even less clear. Furthermore, `::` is commonly used for lists or type hints in Haskell.
> 
> I think a syntax like `as` would be clearer. `(a: A, b: B, c: C) as abc` means we have the destructured tuple `(a,b,c)` but also identify the whole expression as `abc`.

I tried `as` but I think it feels quite clumsy - especially when destructuring arguments of a curried function:
```
.ax %mem.load:  Π .[T: *, .Nat] .as Tas [%mem.M, %mem.Ptr Tas] -> [%mem.M, T], normalize_load;
```
Any other ideas?

leissa:
I think I'm merging for now but add an issue for the alias pattern that we want to have a nicer syntax here. After all, we want to rename Thorin and for this reason I don't want to have stale branches lying around ...

PR #278, opened 2024-10-16
Refactor/pi syntax

leissa:


leissa:
There are still some loose ends that I will fix in later PRs.

PR #294, opened 2025-04-02
allow versioned clang and FileCheck

piotrrak:
This lets to override or use version llvm in lit test-suite.
llvm official ppa provides is as version lets to have multiple versions, while changing symlink is burdensome and unrealible.

### DIFF
--- a/lit/CMakeLists.txt
+++ b/lit/CMakeLists.txt
@@ -7,6 +7,45 @@ else()
   set(LIT_VALGRIND_COMMANDLINE "")
 endif()
 
+find_program(CLANG_EXECUTABLE
+             NAMES clang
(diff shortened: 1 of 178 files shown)


leissa:
nice :)

leissa:
I think we have to put FileCheck in Path somewhere in `macos.yml` and `windows.yml`.

piotrrak:
> I think we have to put FileCheck in Path somewhere in `macos.yml` and `windows.yml`.
It seems we also need to move FileCheck installation before configure. Should be doable will try.

leissa:
superseded by #404

piotrrak:
Thank you!

PR #300, opened 2025-05-05
Python bridge w/ nanobind

leissa:
Python bindings for MimIR.


pytest/regex.py
```diff
@@ -0,0 +1,272 @@
...
+mim.World.call = call
+
+
+class MimRegex:
+    def __init__(self, driver: mim.Driver, *args):
+        self.wrld = driver.world()
+        self.regex = []
+        driver.load_pluins(["core", "compile", "regex", "opt"])
+
+        self._star_sym = self.wrld.sym(f"%regex.quant.star")
```
hsqStephenZhang: `world.call` can receive a str as it's first argument, is it necessary to use `sym` here?

pytest/regex.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def any(self) -> Self:
+        self.regex.append(self.wrld.call(self._any_sym))
+        return self
+
+    def build(self, driver):
+        # print(f"result of annexing with call: {self.wrld.call('%mem.M')}")
+        self.match_func = self.wrld.mut_con(
+            [
+                self.wrld.call("%mem.M", self.wrld.lit_nat_0()),
```
hsqStephenZhang: can these be made as built-in helper functions? they are very common, right?

pytest/regex.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+reg = MimRegex(driver)
+# seg fault
+# reg.any().close()
+reg.literal("a").literal("b").literal("c").literal("d").close().star().close()
+
+print(reg.regex)
+reg.build(driver)
+reg.wrld.dot("out_no_literals", True, False)
```
hsqStephenZhang: can the `reg` be compiled and executed?

mim_py/src/mim_py/__init__.pyi
```diff
@@ -0,0 +1,216 @@
...
+        ...
+class Def:
+    def dump(self) -> None:
+        ...
+    def externalize(self) -> None:
+        ...
+    def num_projs(self) -> int:
+        ...
+    @typing.overload
+    def proj(self, arg0: int, arg1: int) -> Def:
```
hsqStephenZhang: should `__getitem__(self, index)` be also supported? i think it's easier to use. of course, it should be implemented the same as `proj`

mim_py/src/mim_py/__init__.pyi
```diff
@@ -0,0 +1,216 @@
...
+    def sym(self, arg0: str) -> Sym:
+        ...
+    @typing.overload
+    def sym(self, arg0: str) -> Sym:
+        ...
+    @typing.overload
+    def sym(self, arg0: str) -> Sym:
+        ...
+class World:
+    def annex(self, arg0: ...) -> ...:
```
hsqStephenZhang: can we have more detailed typings? IMO ruff/ty/pylance could be much happier with clear types.

src/mim/py/world.cpp
```diff
@@ -0,0 +1,100 @@
...
+        )
+        .def("type_idx", 
+            static_cast<const mim::Def* (World::*)(nat_t)>(&mim::World::type_idx), py::return_value_policy::reference_internal
+        )
+        .def("cn",
+            [](mim::World& w,std::vector<Def*> dom){
+                return w.cn(Defs(dom));
+            }
+        )
+        .def("lit_i8",
```
hsqStephenZhang: lit_ff and other methods for float literal are also helpful.

py/bindings/sym.cpp
```diff
@@ -0,0 +1,27 @@
...
+namespace fe {
+
+void init_sym(py::module_& m) {
+    py::class_<fe::Sym, std::unique_ptr<fe::Sym, py::nodelete>>(m, "Sym")
+        .def(py::init<>())
+        .def("empty", &fe::Sym::empty)
+        .def("size", &fe::Sym::size)
+        .def("view", &fe::Sym::view)
+        .def("str", &fe::Sym::str);
+}
```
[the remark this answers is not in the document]
leissa: @copilot apply changes based on [this feedback](https://github.com/mimir/mimir/pull/300#discussion_r3226542348)

leissa:
I've just ported pybind11 to nanobind which is much more lightweight and reduces build times.

py/src/mim/jit.py
```diff
@@ -0,0 +1,41 @@
...
+        self.ll_name = so_name + ".ll"
+        self.exports = list(exports)
+        self._lib = None
+        self._so_dir = None
+
+    def _get_so_path(self) -> str:
+        system = pf.system()
+        if system == "Windows":
+            if self._so_dir is None:
+                self._so_dir = tempfile.mkdtemp(prefix=f"{self.so_name}-")
```
hsqStephenZhang, line 23: some log here?

.github/workflows/linux.yml
```diff
@@ -42,20 +43,23 @@ jobs:
             -DCMAKE_BUILD_TYPE=${{matrix.build-type}}                                                                                                  \
             -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined,leak,bounds -fsanitize-address-use-after-scope -fno-omit-frame-pointer -fno-sanitize=vptr" \
             -DBUILD_TESTING=ON                                                                                                                         \
+            -DMIM_BUILD_PYTHON=OFF                                                                                                                     \
```
hsqStephenZhang, line 46: why disabled only on linux?

leissa: address sanitizer doesn't play well when running the whole python interpreter. We have it enabled in the Release build, though.

py/src/mim/jit.py
leissa: what do you mean?

hsqStephenZhang (commented):
The stubs are generated in the build dir and packed into the wheel product, but the local development lacks typing support and vscode is unhappy about it, any better ways to address this problem? (i have to copy them manually and add to gitignore as a temporal solution)

py/src/mim/jit.py
hsqStephenZhang: to add some `logging.debug` to the tempdir and compile flags.

leissa:
> The stubs are generated in the build dir and packed into the wheel product, but the local development lacks typing support and vscode is unhappy about it, any better ways to address this problem? (i have to copy them manually and add to gitignore as a temporal solution)

Let's postpone this concern for another PR. Let's merge to master, and then we can discuss where to put the pyi file etc in the new PR.

py/src/mim/plugin.py
```diff
@@ -0,0 +1,36 @@
...
+    def build(self):
+        raise NotImplementedError("Needs to be implemented by plugin author")
+
+    def register_func(self, name: str, input_types: list, return_type):
+        self._registered_functions[name] = (input_types, return_type)
+
+    def jit(self) -> dict[str, MimCallable]:
+        self.build()
+
+        self.driver.world().optimize()
```
hsqStephenZhang, line 26: should this be optional? we could add an `optimize = True` flag.

leissa: right now, pretty much nothing works, if you disable optimization.

JoeJoeSch (approved):

hsqStephenZhang (approved):
LGTM

py/src/mim/jit.py
leissa: I'll leave this for another PR. Would be better to use MimIR's own logging mechanism for this.

PR #308, opened 2025-06-06
Adding get and range functions

lfrenot:
map_range is implemented as an axiom to work around typing issues


leissa:
* I did a bit of formatting
* removed `get_` prefix. I think `first` and `last` is just fine
* There is actually no need for `%vec.range` and `%vec.map_range` as packs can already do that. Check out the updated examples in `lit/vec/normalizers.mim`

PR #312, opened 2025-06-18
Feature/tensor

leissa:


fodinabor:
This is already huge.. should we merge this and iterate in-tree?

leissa:
I will have a more thorough look later today. But one thing that I've noticed: You can try to reduce those chains of `app->decurry()->decurry()` with some variant of  [`App::uncurry`](https://mimir.github.io/classmim_1_1App.html#member-group). If you need to further destruct each curry group, this doesn't buy you much, unfortunately, as C++ doesn't support nested structured bindings.

src/mim/plug/tensor/tensor.mim
```diff
@@ -121,11 +51,241 @@ axm %tensor.reduce: {r: Nat, s: «r; Nat»}
...
+      let bs       = ‹i: nb; s1#(b1#i)›;
+      let bc_1     = %vec.cat (b1, c1);
+      let s1_res   = %vec.diff (s1, bc_1);
+      let n_s1_res = %vec.len s1_res;
+      let bc_2     = %vec.cat (b2, c2);
+      let s2_res   = %vec.diff (s2, bc_2);
+      let n_s2_res = %vec.len s2_res;
+      let s12_res  = %vec.cat (s1_res, s2_res);
+      let s_out    = %vec.cat (bs, s12_res);
+      s_out;
```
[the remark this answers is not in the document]
fodinabor: Any idea why this might have been necessary?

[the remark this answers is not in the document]
leissa: Good question. Looks like a copy-pasta-error to me.

leissa:
One more thing that came to my mind: One reason why I was reluctant to merge some prior version was that importing tensor.mim significantly increases testing times - which happens pretty much everywhere as it is implicitly imported through `compile.mim`. Now, we don't need a solution for this tomorrow, but would be nice to have a solution for this problem in the near term. `make lit` on my machine takes 17 sec vs 60 sec (master vs this branch). This problem is even more severe on our workflows.

src/mim/plug/tensor/phase/fuse.cpp
```diff
@@ -0,0 +1,282 @@
...
+    app->world().DLOG("  subs = {} : {}", subs, subs->type());
+    app->world().DLOG("  comb = {} : {}", comb, comb->type());
+    app->world().DLOG("  init = {} : {}", init, init->type());
+    app->world().DLOG("  Tis = {} : {}", Tis, Tis->type());
+    app->world().DLOG("  Ris = {} : {}", Ris, Ris->type());
+    app->world().DLOG("  Sis = {} : {}", Sis, Sis->type());
+    app->world().DLOG("  To = {} : {}", To, To->type());
+    app->world().DLOG("  Ro = {} : {}", Ro, Ro->type());
+    app->world().DLOG("  nis = {} : {}", nis, nis->type());
+    app->world().DLOG("  is = {} : {}", is, is->type());
```
leissa: just:
```
DLOG(...);
```
The base class provides a `log()` method such that this macro works without getting the world from somewhere.

src/mim/plug/tensor/phase/lower.cpp
```diff
@@ -0,0 +1,133 @@
...
+    auto callee                      = c->as<App>();
+    auto [T, r_in, r_out]            = callee->args<3>();
+    w.DLOG("lower_broadcast_in_dim");
+    w.DLOG("    s_out = {} : {}", s_out, s_out->type());
+    w.DLOG("    input = {} : {}", input, input->type());
+    w.DLOG("    index = {} : {}", index, index->type());
+    w.DLOG("    T = {} : {}", T, T->type());
+    w.DLOG("    r_in = {} : {}", r_in, r_in->type());
+    w.DLOG("    r_out = {} : {}", r_out, r_out->type());
+    w.DLOG("    s_in = {} : {}", s_in, s_in->type());
```
leissa: just
```
DLOG(...);
```

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+    auto callee       = c->as<App>();
+    auto [T, r, s]    = callee->args<3>();
+
+    w.DLOG("lower_get");
+    w.DLOG("    arr = {} : {}", arr, arr->type());
+    if (auto arr_seq = arr->type()->isa<Seq>()) w.DLOG("    arr shape = {}", arr_seq->arity());
+    w.DLOG("    index = {} : {}", index, index->type());
+    w.DLOG("    T = {} : {}", T, T->type());
+    w.DLOG("    r = {} : {}", r, r->type());
+    w.DLOG("    s = {} : {}", s, s->type());
```
leissa: see above

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+    auto size = index->num_projs();
+    w.DLOG("size = {}", size);
```
leissa: see above

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+    auto& w  = new_world();
+    auto c   = rewrite(app->callee());
+    auto arg = rewrite(app->arg());
+
+    auto [arr, index, x] = arg->projs<3>();
+
+    w.DLOG("lower_set");
+    w.DLOG("    arr = {} : {}", arr, arr->type());
+    w.DLOG("    index = {} : {}", index, index->type());
+    w.DLOG("    x = {} : {}", x, x->type());
```
leissa: dito - I'll skip the next instances. just grep over the phase code.

src/mim/plug/tensor/tensor.mim
```diff
@@ -19,96 +19,26 @@ plugin vec;
 let Ring = [
```
leissa: maybe we should make it `%tensor.Ring` such that we can refer to this thing from C++?

src/mim/plug/tensor/tensor.mim
```diff
@@ -121,11 +51,241 @@ axm %tensor.reduce: {r: Nat, s: «r; Nat»}
...
 /// Returns a tensor obtained by folding `f` following the indexes in `subs`
 ///
+/// `nis`, `Ro`, `Ris` and `subs` must be literals, before entering the `tensor::Lower` phase, for the lowering to succeed.
+///
 axm %tensor.map_reduce: {nis: Nat}
                       → {To: *, Ro: Nat}
                       → [So: «Ro; Nat»]
                       → {Tis: «nis; *», Ris: «i: nis; Nat», Sis: «i:nis; «Ris#i; Nat»»}
-                      → [f: [To, «i: nis; Tis#i»] → To, init: To]
+                      → [f: Fn [To, «i: nis; Tis#i»] → To, init: To]
```
leissa, line 60: it's kind of a pity that we already have to decide here for CPS. At that abstraction level I think it would be cooler to have everything still in DS or at least have both options. But I dunno how feasible that is. BTW: I'm currently rewriting the CPS conversion as our current thing is buggy as hell. Maybe this helps?

src/mim/plug/tensor/tensor.mim
```diff
@@ -121,11 +51,241 @@ axm %tensor.reduce: {r: Nat, s: «r; Nat»}
...
+///
+/// ### %%tensor.map_reduce_pure
+///
+/// Variant of %tensor.map_reduce that takes a pure function as input.
+///
+axm %tensor.map_reduce_pure: {nis: Nat}
+                           → {To: *, Ro: Nat}
+                           → [So: «Ro; Nat»]
+                           → {Tis: «nis; *», Ris: «i: nis; Nat», Sis: «i:nis; «Ris#i; Nat»»}
+                           → [f: [To, «i: nis; Tis#i»] → To, init: To]
```
leissa, line 73: oh, Ic. But with "pure" you really mean "ds", no?

src/mim/plug/tensor/phase/fuse.cpp
```diff
@@ -0,0 +1,282 @@
...
+            }
+        } else {
+            auto pos          = new_pos[i];
+            new_Tis_vec[pos]  = Tis->proj(nis_nat, i);
+            new_Ris_vec[pos]  = Ris->proj(nis_nat, i);
+            new_Sis_vec[pos]  = Sis->proj(nis_nat, i);
+            new_subs_vec[pos] = subs->proj(nis_nat, i);
+            new_is_vec[pos]   = is->proj(nis_nat, i);
+        }
+    }
```
leissa, line 183: I'm pretty sure you are mixing old-world with new-world stuff here - which is doomed to explode.
Eg:
```
w.tuple(new_Tis_vec)
```
This thing creates a new-world tuple but `new_Tis_vec` is still from the old world because it stems utlimately from `Tis->proj(nis_nat, i)`, no?

Plz double-check the remaining code below as well. We should probably add an assertion somewhere that immediately fires when constructing a mixed-world node.

src/mim/plug/tensor/phase/fuse.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+
+const Def* Fuse::rewrite_imm_App(const App* app) {
+    if (auto mr = Axm::isa<tensor::map_reduce>(app)) {
+        if (auto res = fuse_map_reduce(mr)) {
+            new_world().DLOG("Fused map_reduce at {} into a new map_reduce {}", app, res);
+            return res;
+        }
+    }
+    return Rewriter::rewrite_imm_App(app);
```
leissa: I recently switched on `master` to
```cpp
return RWPhase::rewrite_imm_App(app);
```
which is arguably a bit more idiomatic as `RWPhase` is the direct super class.

src/mim/plug/tensor/phase/lower.cpp
```diff
@@ -0,0 +1,133 @@
...
+    // chain bottoms out at the low-level axioms (`map_reduce`, …) in one go.
+    return impl;
+}
+
+const Def* Lower::lower_broadcast_in_dim(const App* app) {
+    auto& w  = new_world();
+    auto c   = rewrite(app->callee());
+    auto arg = rewrite(app->arg());
+
+    auto [s_in, s_out, input, index] = arg->projs<4>();
```
leissa, line 43: I'm pretty sure you are again mixing old-world with new-world nodes here. `arg->projs...` is old-world stuff that has not been rewritten/carried over yet.

src/mim/plug/tensor/phase/lower.cpp
```diff
@@ -0,0 +1,133 @@
...
+    w.DLOG("    r_in = {} : {}", r_in, r_in->type());
+    w.DLOG("    r_out = {} : {}", r_out, r_out->type());
+    w.DLOG("    s_in = {} : {}", s_in, s_in->type());
+
+    auto r_in_lit = r_in->isa<Lit>();
+    if (!r_in_lit) return nullptr;
+    auto r_in_nat  = r_in_lit->get<u64>();
+    auto r_out_lit = r_out->isa<Lit>();
+    if (!r_out_lit) return nullptr;
+    auto r_out_nat = r_out_lit->get<u64>();
```
leissa: you can shortcut a bit. Sth like this:
```c++
auto r_in_nat  = Lit::isa(r_in);
auto r_out_nat = Lit::isa(r_out);
if (!r_in_nat || !r_out_nat) return nullptr;
```
This will give you an `std::optional<nat_t>` (with `nat_t` being the default template arg).

src/mim/plug/tensor/phase/lower.cpp
```diff
@@ -0,0 +1,133 @@
...
+        auto idx     = index->proj(r_in_nat, i);
+        auto idx_lit = Lit::isa(idx);
+        if (!idx_lit) return nullptr;
+        u64 idx_nat = *idx_lit;
+
+        map_perm[idx_nat] = i;
+
+        set_perm.erase(idx_nat);
+    }
+    u64 j = r_in_nat;
```
leissa: I know copilot complained to prefer using absl, but are you sure that unordered set doesn't break this code here?

src/mim/plug/tensor/phase/lower.cpp
```diff
@@ -0,0 +1,133 @@
...
+    for (auto i = set_perm.begin(); i != set_perm.end(); i++) {
```
leissa: If you want to be cool:
```cpp
for (u64 j = r_in_nat; auto i : set_perm)
    map_perm[i] = j++;
```

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+    w.DLOG("lower_get");
+    w.DLOG("    arr = {} : {}", arr, arr->type());
+    if (auto arr_seq = arr->type()->isa<Seq>()) w.DLOG("    arr shape = {}", arr_seq->arity());
+    w.DLOG("    index = {} : {}", index, index->type());
+    w.DLOG("    T = {} : {}", T, T->type());
+    w.DLOG("    r = {} : {}", r, r->type());
+    w.DLOG("    s = {} : {}", s, s->type());
+
+    auto size = index->num_projs();
```
leissa: This is usually a bad idea - unless you know what you are doing. Better got get the number of supposed elems elsewhere. `num_projs` can always give you trouble if the outer thing degenerates to a 1-tuple that contains a 3-tuple or so. But I don't know your code well. Maybe it's fine here.

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+    w.DLOG("size = {}", size);
+    if (size == 1) {
+        w.DLOG("index of size 1, extract");
+        return w.extract(arr, index);
+    }
+
+    auto r_lit = r->isa<Lit>();
```
leissa: once again `Lit::isa(r)`

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -0,0 +1,552 @@
...
+                dims[idx_nat] = dim;
+                w.DLOG("        {} ↦ {}", idx_nat, dim);
+            } else {
+                auto prev_dim = dims[idx_nat];
+                w.DLOG("        prev dim {} = {}", idx_nat, prev_dim);
+                // override with more precise information
+                if (auto dim_lit = dim->isa<Lit>()) {
+                    if (auto prev_dim_lit = prev_dim->isa<Lit>()) {
+                        if (dim != prev_dim) {
+                            if (!dim_lit) throw std::runtime_error("dimension is not a literal");
```
leissa: better use:
```
error(fmt_string)
```
it will also throw but gives you a fmt string and we have a unique interface to report errors. A unique interface will also help us supporting exceptions through FFI such as our new python bindings.

leissa:
Footnote: As I'm officially author of this PR, I can't approve this PR. Just feel free to merge, whenever you think it looks good :)

Looks already pretty good to me :+1: 

Tagging @HarithF here, as he might want to have a look as well.

HarithF:
I am working on a branch of this for the MLIR emitter, so it looks good to me as well.
But I will take another look tomorrow.

src/mim/plug/tensor/tensor.mim
fodinabor: Could do.. but it really is just used to the define the type.. unless we actually want to construct new axioms from C++, I don't see the need..? 🤔

src/mim/plug/tensor/phase/fuse.cpp
fodinabor: And `Tis` should already be rewritten thanks to L34.
+1 for the assertion :D

src/mim/plug/tensor/phase/lower.cpp
fodinabor: same as above.. shouldn't `rewrite(app->arg())` on a tuple give us already rewritten stuff?

src/mim/plug/tensor/tensor.mim
fodinabor: When wanting to be explicit, it's probably nice
```rust
let R: %tensor.Ring = (Nat, 0, %core.nat.add, %core.nat.mul);

```

fodinabor:
I'd be happy to merge once the CI is done, unless @HarithF has further comments.

fodinabor (approved):
Let's iterate in-tree.

PR #313, opened 2025-07-09
[Affine] Infer type arguments for `affine.For`.

fodinabor:
No reason to make it harder than it has to be to use `affine.For`.


fodinabor:
I've included a fix for extracting from a 1-sized tuple (which is normalized away), where the size is inferred.
This might be a bit ad-hoc / at the wrong place, since I'm not too deep into the new zonking stuff.

leissa:
Found the issue with the type inference. There were actually some dependencies missing for the new free_vars magic which caused the Zonker to not descend into some sub-expressions in some edge cases. See #314

PR #317, opened 2025-07-20
Rewrites

leissa:


src/mim/ast/emit.cpp
```diff
@@ -497,4 +497,13 @@ void CDecl::emit(Emitter& e) const {
     }
 }
 
+void RuleDecl::emit(Emitter& e) const {
+    auto _      = e.world().push(loc());
+    auto meta_t = e.world().rule_type(var()->emit_type(e));
+    auto rule_  = e.world().mut_rule(meta_t);
+    var()->emit_value(e, rule_->var());
+    const Rule* res = rule_->set(lhs()->emit(e), rhs()->emit(e), condition()->emit(e));
```
leissa, line 505: order of eval for function calls is impl-defined in C++. This has the unfortunate effect that on some C++ impls we may experience slightly different stuff happening here. Better to split it like this:
```c++
auto l = lhs()->emit(e);
auto r = rhs()->emit(e);
auto c = condition()->emit(e);
auto res = rules_->set(l, r, c);
```

src/mim/ast/lexer.cpp
```diff
@@ -59,6 +59,7 @@ Tok Lexer::lex() {
         if (accept(U'⟩')) return tok(Tag::D_angle_r);
         if (accept( '<')) {
             if (accept( '<')) return tok(Tag::D_quote_l);
+            if (accept('=') && accept('>')) return tok(Tag::T_equiv);
```
leissa, line 62: If we get `<=` we run into trouble, no?
Better:
```c++
if (accept('<'))
    if (accept( '<')) return tok(Tag::D_quote_l);
    if (accept('='))
        if (accept('>')) return tok(Tag::T_equiv);
    // error: illegal token
```

src/mim/rule.cpp
```diff
@@ -0,0 +1,122 @@
...
+        return true;
+    }
+    return false;
+}
+
+const Def* Rule::replace(const Def* expr, std::map<const Def*, const Def*>& v2v) const {
+    // the dic is now populated with everything
+    std::vector<std::pair<const Def*, size_t>> tuple_of_args;
+    const Var* var_of_rule;
+    for (auto v : v2v) {
```
leissa, line 98:
```c++
for (auto [var, val] : v2v)
```

PR #358, opened 2025-12-11
Refactor/extendable ll emitter

ftschirpke:
This PR allows the LLVM emitter to be extended.
Additionally, this PR adds virtual functions to the emitter that I have found to be necessary/useful while implementing a prototype that emits LLVM IR for host code and device code separately.

Unfortunately, I have not found a way to link a new plugin against the `core` plugin to actually use this.
Thus, I want to ask for help as I do not see a solution, short of moving the full implementation of the emitter from `ll.cpp` to the header `ll.h` or including `ll.cpp` in the new plugin's source files again.


ftschirpke:
As discussed, I moved the LL emitter implementation to the header as a first and simple solution.
Furthermore, I added three points of extensibility that should suffice (at least they do for my GPU use cases right now):
- `virtual std::optional<std::string> isa_targetspecific_intrinsic(BB&, const Def*)` - the primary extension point - a function that is called when the LL emitter could match nothing else 
- `virtual std::string prepare()` - is now virtual s.t. extending class can control how function arguments are emitted
  - GPU use case: add label to kernels and modify kernel arguments for block index and thread index
- `virtual void emit_epilogue(Lam*)` - is now virtual - this might not be necessary, but currently it is needed:
  - GPU use case: if `%gpu.launch (..) (..) $ (..)` is the top-level application in the function body i.e. is `lam->body()`, then the current LL emitter will never try to emit and match it as the whole application but only with the last application removed. This makes it impossible for me to match it with `if (auto launch = Axm::isa<gpu::launch) ...` and thus I extend the `emit_epilogue` function with a check catching exactly this edge case.

PR #387, opened 2026-04-11
Option type

HarithF:
WIP for option type {[] ∪ T;}
Plugin itself complete, integration with org.get still in progress. normalize_get does not return option yet, therefore 2 tests fail.


HarithF:
Option type {[] ∪ T;}
supported safe and unsafe unwrap operations for some and none

leissa:
:+1:

PR #395, opened 2026-04-20
Feature/eqsat

ashiven:
This PR adjusts the `slotted-egraphs` part of the symbolic expression emitter for the out-of-tree [eqsat](https://github.com/ashiven/eqsat) plugin.

It also slightly changes the definition of `Rule` to make its meta variables accessible.


include/mim/rule.h
```diff
@@ -37,19 +37,21 @@ class Reform : public Def, public Setters<Reform> {
...
 
 public:
     /// @name lhs & rhs
     /// @see @ref proj
     ///@{
     const Reform* type() const { return Def::type()->as<Reform>(); }
-    const Def* lhs() const { return op(0); }
-    const Def* rhs() const { return op(1); }
-    const Def* guard() const { return op(2); }
+    const Def* meta_var() const { return op(0); }
```
leissa: Is the `meta_var` a real `Var`? If so (what I think it is), don't make `meta_var` an op.

valgrind.supp
```diff
@@ -0,0 +1,32 @@
...
+  fun:*symbol_table*SymbolTable*intern*
+}
+{
+  egg_symbol_table_symbol_table_realloc
+  Memcheck:Leak
+  match-leak-kinds: possible,reachable
+  fun:realloc
+  ...
+  fun:*symbol_table*SymbolTable*intern*
+}
```
leissa: can we remove this file?

src/mim/plug/core/be/sexpr.cpp
```diff
@@ -212,86 +273,148 @@ void Emitter::finalize() {
...
-void Emitter::emit_lam(Lam* lam, MutSet& done) {
-    done.emplace(lam);
+bool Emitter::is_recursive(Lam* lam) {
+    std::set<Lam*> visited;
+    for (auto next_lam : next_lams(lam)) {
+        if (next_lam == lam) return true;
+        if (is_reachable(lam, next_lam, visited)) return true;
+    }
+
+    return false;
```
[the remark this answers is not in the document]
leissa: We have a much better magic to figure out whether sth is recurse. Plz use the [nesting tree](https://mimir.github.io/classmim_1_1Nest.html#a49652caf6c6f54036e31936e80161753) instead.

leissa (changes requested):
I think the biggest issues are:
* `Rule::meta_var` should probably be a real `Var`
* Use the nesting tree to detect recursion.

AI also has some complaints.

valgrind.supp
ashiven: This was needed because egg uses some global symbol table that created some false positives for memory leaks but I am not sure if its possible to override the existing `VALGRIND_COMMAND_CONFIG` from my own `CMakeLists.txt` in the out-of-tree plugin..

leissa: Sth along those lines should work:
```cmake
set(VALGRIND_COMMAND_CONFIG
    "--tool=memcheck --leak-check=full --show-leak-kinds=all --suppressions=${CMAKE_SOURCE_DIR}/valgrind.supp"
)
```

leissa:
~~The only thing missing is removal of `valgrind.supp` and then we a re ready to merge?~~

leissa (approved):

PR #400, opened 2026-04-24
Don't emit the body for 0-sized arrays.

fodinabor:
One can write something like `<i: nb; s#(b#i)>`, where `b = ()` and `nb` is the size of `b`, i.e. 0. This seems like it could be useful/legal, as we don't really care about the body, when the array is empty anyways...

I am somewhat assuming that there might be more places where this might strike.

### DIFF
--- a/src/mim/ast/emit.cpp
+++ b/src/mim/ast/emit.cpp
@@ -324,24 +324,18 @@ const Def* TupleExpr::emit_(Emitter& e) const {
 
 const Def* SeqExpr::emit_(Emitter& e) const {
     auto s = arity()->emit_type(e);
-    if (auto lit_s = Lit::isa(s); lit_s && *lit_s == 0) return e.world().unit(true);
+    if (auto lit_s = Lit::isa(s); lit_s && *lit_s == 0) return e.world().unit(is_pack());
 
     if (arity()->dbg().is_anon()) { // immutable
         auto b = body()->emit(e);
-        return is_arr() ? e.world().arr(s, b) : e.world().pack(s, b);
+        return e.world().seq(is_pack(), s, b);
     }
 
     auto t = e.world().type_infer_univ();
     auto a = e.world().mut_arr(t);
     a->set_arity(s);
 
-    if (is_arr()) {
-        auto var = a->var();
-        arity()->emit_value(e, var);
-        a->set_body(body()->emit(e));
-        if (auto imm = a->immutabilize()) return imm;
-        return a;
-    } else {
+    if (is_pack()) {
         auto p   = e.world().mut_pack(a);
         auto var = p->var();
         arity()->emit_value(e, var);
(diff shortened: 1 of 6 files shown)


leissa:
So, multi-indices is currently WIP.

The main thing which is still missing is `a#is` where `is` is an index-pack/tuple. 

For this reason:
```
((a, b), (c, d), ,(c, d))#(1_3, 0_2) => ((a, b), (c, d), ,(c, d))#1_3#0_2
```
This already happens via normalization. We also consistently transform multi-dim arrays and packs (#399):
```
«(i, j, k); body» => «i, «j, «k; body»»»
```
This mirrors the normalization above. What *is* missing though, is support for multi-indices whose arity is parametric such as this:
```rust
let is:   «n; Idx 3» = /*...*/;
let as:   «n; T»     = /*...*/;
let elem: T          = as#is;
```
And I'm currently thinking that `n` can be`0` as in  «0; T». For this reason, we should also allow extraction with an empty index-tuple `()`. No? In other wordds `()` is the neutral element of extraction.

fodinabor:
I generally agree with your argument that `()#()` -> `()` (as `<<0;T>>` imo is just `()`)

I believe that this is orthogonal to this PR, however.
In this case, we have `(1,2,3)#(()#i)` where the index is `()#i`.
Since `()` is indexed here first, we would need `()#i` to be valid, but that gives me a `⊥:*`.

The error that I get is the following: `'()#i_10619' is not of Idx type but of type '⊥:*'`
So the problem is not that we are trying to index into a multi-dim array with a tuple, but instead that the tuple we try to get the index from is empty.

Full abstract example:
```
let nb = 0;
let b = ();

let s1 = (2, 2);

let bs = ‹i: nb; s1#(b#i)›;
```

Concrete usecase in the tensor dialect - we can have a list of batching dimensions for `tensor.dot_product`. However, for e.g. the normal 2d product case, we don't have any of those, therefore [product_2d](https://github.com/AnyDSL/MimIR/blob/cc4f7d5002f03cb3c3d874cd6ce585dca6bc067d/src/mim/plug/tensor/tensor.mim#L152) provides an empty tuple, which then blows up here:
https://github.com/AnyDSL/MimIR/blob/cc4f7d5002f03cb3c3d874cd6ce585dca6bc067d/src/mim/plug/tensor/tensor.mim#L105

This PR tackles this issue by not emitting/rewriting the body of `‹i: nb; s1#(b#i)›` if `nb` is `0`.

leissa:
Ah yes, you are absolutely right. The same reason why we need "smart rewriting" in the `Rewriter`. If we have `‹n; body›`, and `n` turns out to be `0`, we must **not** rewrite `body`, otherwise we see inconsistencies. Took me ages until I finally realized this... And the same is true, even if we emit stuff.

src/mim/ast/emit.cpp
```diff
@@ -324,6 +324,8 @@ const Def* TupleExpr::emit_(Emitter& e) const {
 
 const Def* SeqExpr::emit_(Emitter& e) const {
     auto s = arity()->emit_type(e);
+    if (auto lit_s = Lit::isa(s); lit_s && *lit_s == 0) return e.world().unit(true);
+
```
leissa, line 328: This is not entirely correct, though. `SeqExpr` is either a pack or an arr. We have to pass as flag to `world().unit(is_pack)` to either build a pack or an array. While at it, I saw that the logic for `SeqExpr` and the `world` methodd was exactly flipped. So while at it, I flipped `SeqExpr::is_arr()` to `SeqExpr::is_pack()` to have it consistent with the seq-style methods in `World`.

PR #406, opened 2026-05-05
Bugfix/add_mim_plugin

ashiven:
If `add_mim_plugin` is run inside of `extra/plugin/CMakeLists.txt` the plugin search path for bootstrapping `autogen.h` expands to `mimir/extra` instead of `mimir/mim/src/plug` and an extra plugin can't import any in-tree plugins.


leissa (approved):
lgtm 👍

PR #417, opened 2026-05-15
fix(py): return lam instead of def in lam.set

hsqStephenZhang:
## Background

`Lam`'s set method is inherited from `Def`

So after calling `.set` for a `Lam` in python, the result will be inferred as a `Def` and is not callable at type level.

## Solution

overwrite the method for `Lam` so nanobind will generate correct stubs.

### DIFF
--- a/py/bindings/lam.cpp
+++ b/py/bindings/lam.cpp
@@ -1,4 +1,5 @@
 #include <nanobind/nanobind.h>
+#include <nanobind/stl/string.h>
 #include <nanobind/stl/vector.h>
 
 #include <mim/def.h>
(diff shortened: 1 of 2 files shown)


hsqStephenZhang:
the windows CI failure seems unrelated with this fix.

leissa:
I guess there is sth wrong with how `std::string` is passed around - causing the failing test on Windows.

leissa:
I think you are right. it's unrelated. some weird ordering issue in test discovery or sth like that.

leissa:
skipping the test on Win doesn't even fix this, as the test discovery itself has some ordering issues

leissa:
AI suggest this fix (can't test right now):
```diff
index 0c5f7d6..0000000 100644
--- a/py/tests/call.py
+++ b/py/tests/call.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import pytest
 
 import mim
-import mim.plug.core as core
 
 
 @pytest.fixture
@@ -15,6 +14,8 @@ def core_world(driver, plugin_dir):
 
 
 def test(core_world):
+    import mim.plug.core as core
+
     assert isinstance(core_world.call(core.bit2.and_), mim.Def)
     nat0 = core_world.lit_nat_0()
     assert isinstance(core_world.call(core.bit2.and_, nat0), mim.Def)
```

leissa:
I now have in fix in #430. Problem was that the `set` method was actually injected through a mixin which causes some hicups on msvc. I'm now using a lambda.

PR #418, opened 2026-05-15
Move broadcast_in_dim into lowering phase as well.

HarithF:
Nothing can be normalized in that case afaik


fodinabor:
LGTM, thx!

PR #420, opened 2026-05-16
feat: fold runtime vector

hsqStephenZhang:
## Background

before this fix, the `normalize_fold` cannot handle runtime vectors, thus it will throw an unhanded def error in llvm:

```bash
./build/bin/mim ../MimIR/lit/vec/fold_runtime.mim --output-ll - # the fold_runtime.mim is added as a test case
error: unhandled def in LLVM backend: s_31307 : %math.F (52, 11)
```

## Solution
 
- Extend normalize_fold to handle runtime arrays: when the vector has a compile-time-known arity, scalarize via
  element-wise Extract + app instead of returning nullptr.
 - Fix the Extract handler in ll.h: for SIMD-typed tuples with a literal index, emit extractelement instead of returning
  the whole vector register.

### DIFF
--- a/include/mim/plug/core/be/ll.h
+++ b/include/mim/plug/core/be/ll.h
@@ -660,7 +660,14 @@ inline std::string Emitter::emit_bb(BB& bb, const Def* def) {
         auto tuple = extract->tuple();
         auto index = extract->index();
         auto v_tup = emit_unsafe(tuple);
-        if (is_simd(tuple->type()) && !Axm::isa<mem::M>(tuple->type())) return v_tup;
+
+        if (is_simd(tuple->type()) && !Axm::isa<mem::M>(tuple->type())) {
+            if (auto li = Lit::isa(index)) {
+                auto t_vec  = convert(tuple->type(), true);
+                return bb.assign(name, "extractelement {} {}, i32 {}", t_vec, v_tup, *li);
+            }
+            return v_tup;
+        }
 
         // this exact location is important: after emitting the tuple -> ordering of mem ops
         // before emitting the index, as it might be a weird value for mem vars.
--- a/lit/vec/fold_runtime.mim
+++ b/lit/vec/fold_runtime.mim
@@ -13,10 +13,3 @@ plugin vec;
 fun extern fold_l_runtime (v: «4; %math.F64»): %math.F64 =
     let s = %vec.fold.l (%math.arith.add 0) (0.0 : %math.F64, v);
     return s;
-
-// CHECK-LABEL: @fold_r_runtime(
-// CHECK:         extractelement <4 x double>
-// CHECK:         fadd double
-fun extern fold_r_runtime (v: «4; %math.F64»): %math.F64 =
-    let s = %vec.fold.r (%math.arith.add 0) (v, 0.0 : %math.F64);
-    return s;
--- a/lit/vec/fold_runtime.mim
+++ b/lit/vec/fold_runtime.mim
@@ -1,0 +1,22 @@
+// RUN: %mim %s --output-ll - | %FileCheck %s
+//
+// Test: normalize_fold should work on runtime SIMD arrays («n; T»).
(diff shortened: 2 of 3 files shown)


leissa:
Actually, we don't want to scalarize here. We want to use llvm reductions:
https://llvm.org/docs/LangRef.html#vector-reduction-intrinsics

hsqStephenZhang:
That's my plan after this fix. Should I just include it in this PR?

leissa:
I can either close this one w/o merging or you continue working on this branch.

hsqStephenZhang:
ok, i'll work on this branch.

hsqStephenZhang:
> Actually, we don't want to scalarize here. We want to use llvm reductions: https://llvm.org/docs/LangRef.html#vector-reduction-intrinsics

I found that the reduce semantics looks similar to the fold but not entirely, the main issue is that

not all reduce intrinsics have start value. for integers, the add/mul/and/... has only a vector argument, fadd/fmul has start value, fmax/fmin does not have start value either. 

i think it's just too much for a single `%vec.fold` axiom. we might be able to optimize the %vec.fold into vectorized reduce in some situations, but to use `llvm.vector.reduce` explicitly, I think we need other axioms like `%vec.reduce`.

and i suggest adding two additional axioms `%vec.reduce_acc` and `%vec.reduce`, where the later does not take a start value. combined they could cover the `llvm.vector.reduce.*` intrinsics

hsqStephenZhang:
https://mlir.llvm.org/docs/Dialects/Vector/#vectorreduction-vectorreductionop

MLIR's `vector.reduction` provides a more unified interface though. and the initial_value/acc is optional.

fodinabor:
> not all reduce intrinsics have start value. for integers, the add/mul/and/... has only a vector argument, fadd/fmul has start value, fmax/fmin does not have start value either.

IMO, that is not really an issue, as in case you want to use a start value, you should be able to just apply the operation of the start value at the end as well, when you already have the scalar result, e.g.: 
```llvm
declare i32 @llvm.vector.reduce.add.v4i32(<4 x i32> %a)

%reduced = @llvm.vector.reduce.add.v4i32(<4 x i32> %a)
%final = add i32 %reduced, %start_value
```

Our goal is not really to faithfully model any LLVM (or MLIR) intrinsics to have a 1:1 mapping... we do strive to have consistent, useful semantics in the MimIR world and then try to lower the MimIR world into something efficient in the backend.

leissa:
If we are adding reduction variants w/o initial value, we always have to ask the question what is going to happen, if the length of the involved vector happens to be zero. And then, the result becomes undefined.

fodinabor:
Hmmm... actually, I'd like to revisit this.. it's not so much about scalarization in the `normalize_fold`, it's more about enabling compile-time expansion of the fold.

E.g. I'd be interested in using `vec.fold.l` in such a context:
```rust
/// Accumulator for the axis-sum: adds `row#ax` to the running total.
lam concat_acc {r: Nat} (ax: Idx r) (acc: Nat, row: «r; Nat»): Nat = %core.nat.add (acc, row#ax);

/// `(ax, Sis) ↦ s_out`: sums the inputs' extents along `ax`, keeps the shared extents on every other axis.
lam concat_shape {nis r: Nat} (ax: Idx r, Sis: «nis; «r; Nat»»): «r; Nat» =
    let sum_ax = %vec.fold.l (concat_acc ax) (0, Sis);
    let row_0  = Sis#(%core.idx nis 0 0);
    ‹d: r; %core.select (%core.icmp.e @r (d, ax), sum_ax, row_0#d)›;

axm %tensor.concat: {T: *, nis r: Nat}
                  → [ax: Idx r]
                  → {Sis: «i: nis; «r; Nat»»}
                  → [is: «i: nis; «Sis#i; T»»]
                  → «concat_shape (ax, Sis); T», normalize_concat;
```
Otherwise I'd have to add an `s_out` for limited good reasons... :)

To minimize the SIMD issues, we could restrict the normalization to `Nat` folds, for example?

PR #421, opened 2026-05-18
Refactor/search path2

leissa:
supersedes #419


leissa:
Not perfect but better what we had beforehand. Smarter version is on the way.

PR #422, opened 2026-05-18
testing automatic creation of python bindings

JoeJoeSch:
Python bindings are now automatically created with the help of a python script at build time - woven into cmake.


leissa:
Tagging @fodinabor and @hsqStephenZhang again as I really want your input for this PR.

hsqStephenZhang:
Some C++ implementation types leak into the generated _mim.pyi. For example:

```python
class Sym:
    def rbegin(self) -> "std::__1::reverse_iterator<char const*>": ...
    def rend(self) -> "std::__1::reverse_iterator<char const*>": ...
    def crbegin(self) -> "std::__1::reverse_iterator<char const*>": ...
```

The existing pytest did not catch this because these methods are not invoked. Depending on the type, we should either include the appropriate nanobind STL caster header (for example, pair.h, tuple.h, or optional.h) or exclude unsupported C++-specific APIs using [skip:XXX] in the corresponding .nbextra file.

It would also be helpful to add a Pyright check to CI, together with a test that rejects generated stubs containing leaked C++ type names such as std::, mim::, fe::, or absl::.

hsqStephenZhang:
suggested by codex: nanobind support post-build rewrite by option `-p`, see https://nanobind.readthedocs.io/en/latest/typing.html#pattern-files for more details. Basically, it allows user to add some APIs to the generated stub, here is a workable pattern file:

```txt
# Python APIs added to nanobind classes at package import time, plus signature
# refinements that nanobind cannot infer from overloaded C++ bindings.

^_mim\.Def\.__getitem__$:
    \from typing import overload
    @overload
    def __getitem__(self, index: int, /) -> Def: ...
    @overload
    def __getitem__(self, index: tuple[int, int], /) -> Def: ...

^_mim\.Hole\.find$:
    def find(self) -> tuple[Hole, Def | None]: ...

^_mim\.Axm\.get$:
    def get(arg: Def, /) -> tuple[Axm | None, int, int]: ...

^_mim\.Driver\.add_search_path$:
    \from os import PathLike
    def add_search_path(self, path: str | PathLike[str], /) -> None: ...

^_mim\.World\.__suffix__$:
    \from enum import IntEnum
    def call(
        self,
        callee: Def | Sym | str | IntEnum,
        *args: Def | list[Def] | tuple[Def, ...],
    ) -> Def: ...
```

The benefit of using a nanobind stubgen pattern file is that we no longer need artificial wrapper subclasses such as `Def` and `World` in `py/src/mim/__init__.pyi`. The pattern injects the monkey-patched methods and refined signatures directly into the generated `_mim.pyi`.

Not so relevant to this PR but it would be a nice change.

leissa:
> Some C++ implementation types leak into the generated _mim.pyi. For example:
> 
> ```python
> class Sym:
>     def rbegin(self) -> "std::__1::reverse_iterator<char const*>": ...
>     def rend(self) -> "std::__1::reverse_iterator<char const*>": ...
>     def crbegin(self) -> "std::__1::reverse_iterator<char const*>": ...
> ```
> 
> The existing pytest did not catch this because these methods are not invoked. Depending on the type, we should either include the appropriate nanobind STL caster header (for example, pair.h, tuple.h, or optional.h) or exclude unsupported C++-specific APIs using [skip:XXX] in the corresponding .nbextra file.
> 
> It would also be helpful to add a Pyright check to CI, together with a test that rejects generated stubs containing leaked C++ type names such as std::, mim::, fe::, or absl::.

This is now resolved

leissa:
> suggested by codex: nanobind support post-build rewrite by option `-p`, see https://nanobind.readthedocs.io/en/latest/typing.html#pattern-files for more details. Basically, it allows user to add some APIs to the generated stub, here is a workable pattern file:
> 
> ```
> # Python APIs added to nanobind classes at package import time, plus signature
> # refinements that nanobind cannot infer from overloaded C++ bindings.
> 
> ^_mim\.Def\.__getitem__$:
>     \from typing import overload
>     @overload
>     def __getitem__(self, index: int, /) -> Def: ...
>     @overload
>     def __getitem__(self, index: tuple[int, int], /) -> Def: ...
> 
> ^_mim\.Hole\.find$:
>     def find(self) -> tuple[Hole, Def | None]: ...
> 
> ^_mim\.Axm\.get$:
>     def get(arg: Def, /) -> tuple[Axm | None, int, int]: ...
> 
> ^_mim\.Driver\.add_search_path$:
>     \from os import PathLike
>     def add_search_path(self, path: str | PathLike[str], /) -> None: ...
> 
> ^_mim\.World\.__suffix__$:
>     \from enum import IntEnum
>     def call(
>         self,
>         callee: Def | Sym | str | IntEnum,
>         *args: Def | list[Def] | tuple[Def, ...],
>     ) -> Def: ...
> ```
> 
> The benefit of using a nanobind stubgen pattern file is that we no longer need artificial wrapper subclasses such as `Def` and `World` in `py/src/mim/__init__.pyi`. The pattern injects the monkey-patched methods and refined signatures directly into the generated `_mim.pyi`.
> 
> Not so relevant to this PR but it would be a nice change.

as you mention, this is sth for another PR.

leissa:
alright. merging. nobody objected :)

PR #425, opened 2026-05-20
Rewrite Pack var with Array var when building type

### LINKED ISSUE #414: [bug] lams returning dependent Packs in Tuples are not closed
### Reproducer

```sh
./scripts/checkout.sh git@github.com:mimir/mimir.git 2f759f8e436f968810e198b817304249efab115a && ./scripts/probe.sh lit/bug/pack_not_closed.mim
```

```mim
lam extern test {n: Nat} {Ts: «n; ★»} [values: «i: n; Ts#i»]
  = ‹i: n; values#i›;
```

### Observed behavior

error message:
```
mim: /home/sympel/development/projects/uni/mimir/MimIR/src/mim/world.cpp:29: void mim::World::Externals::externalize(Def *): Assertion `def->is_closed()' failed.
/nix/store/1044bqz6ylvzh6msnq6ybjpvg3hds7bg-mim/bin/mim: line 40: 296838 Aborted                    (core dumped) $DBG $MIM_BIN -P build/lib64/mim "$@"
```

The issue seems to be that the type of the Pack (an Array) reuses the Pack's var, leading to it appearing freely there and then transitively in the lams Pi type and therefore in the lam itself.

### Expected behavior

No output.

I would expect the Array to introduce its own var and use that instead of reusing the Packs var.

amaebel:
fixes #414

When emitting a Pack from mim, check whether it has a Var and rewrite it with the Arr's var if so before setting the Arr's body. This prevents the Pack's var from being free in its type, which caused any expression containing dependent Packs to not be closed, as in the linked issue.

### DIFF
--- a/lit/bug/pack_not_closed.mim
+++ b/lit/bug/pack_not_closed.mim
@@ -1,3 +1,5 @@
+// RUN: %mim %s -o -
+
 lam extern test {n: Nat} {Ts: «n; ★»} [values: «i: n; Ts#i»]
   : [Nat, «i: n; Ts#i»]
   = (0, ‹i: n; values#i›);
--- a/lit/bug/pack_not_closed.mim
+++ b/lit/bug/pack_not_closed.mim
@@ -1,0 +1,2 @@
+lam extern test {n: Nat} {Ts: «n; ★»} [values: «i: n; Ts#i»]
+  = ‹i: n; values#i›;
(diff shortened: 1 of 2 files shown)


amaebel:
I *think* the problem with the test is that dependent packs are still broken with optimizations. I could just disable those for the tests, but that seems questionable, as this is obviously also a bug. Although merging failing tests onto master is probably also wrong.

The options are:
- add `-O 1` to test cases for now
- don't merge until other bug fixed
- merge failing test and make CI ignore it for now

the same applies to #427

leissa:
There are a ton of things that we can't handle yet in our code generator etc. Just use `-O0` in the lit script or sth like that.

leissa (approved):
lgtm

PR #427, opened 2026-05-20
Add %tuple.head and %tuple.tail

amaebel:
This PR adds the lams %tuple.head and %tuple.tail, which are frequently useful when implementing recursive functions consuming tuples. It does currently not check that the tuple is not empty, as that broke type checking in some of my use cases, so the error messages for n=0 are not great. 

Depends on #425


leissa (approved):
Just fix the style issue at pointed out by copilot, then we are ready to merge.

PR #428, opened 2026-05-20
Add axioms to inspect closedness of expressions to refly

amaebel:
This PR introduces two new axioms to refly that allow to inspect expression closed-ness:
`%refly.vars.is_closed e`: Checks whether the expression `e` is closed. In contrast to `%core.pe.is_closed`, this always normalizes immediately upon construction to either `tt` or `ff`.
`%refly.vars.await_closed guard e`: Normalizes to expression `e` once expression `guard` is closed. This allows precise control over when other normalers are fired depending on other expressions becoming constant. Combined with `%refly.vars.is_closed`, it allows for example to check whether some expression is still not closed once a different expression is, which can be used to delay checking closed-ness inside a lam body until it gets reduced/partially evaluated.

Notably, `%core.pe.is_closed e` could now be rewritten as `%core.refly.await_closed e tt`, if that is desired.

### DIFF
--- a/src/mim/plug/refly/normalizers.cpp
+++ b/src/mim/plug/refly/normalizers.cpp
@@ -95,6 +95,25 @@ const Def* normalize_check(const Def* type, const Def*, const Def* arg) {
     return nullptr;
 }
 
+template<vars id>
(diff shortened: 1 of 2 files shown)


leissa (approved):
Lgmt

PR #433, opened 2026-06-01
Add assertion to check if all ops of a Def are in the current world.

fodinabor:
As brought up in #312 

For most tests that I tried to fabricate, it crashed somewhere earlier already (due to e.g. Zonk-ing), however this should still capture otherwise silent issues.


leissa (approved):
lgtme

PR #435, opened 2026-06-03
more idiomatic use of std::views

leissa:
@fodinabor : can you have a look at [src/mim/plug/regex/normalizers.cpp](https://github.com/mimir/mimir/pull/435/changes#diff-dba4473830aa2182eee4d2e13a5dffb35ab6aeb77d84df5ec01f3e590fe794f2)? AI helped me with that.


fodinabor (approved):
LGTM.

PR #437, opened 2026-06-05
[core] Implement identity normalization for double conversion in normalize_conv

fodinabor:
This only applies if the first conv is a widening conversion and the second conversion converts back to the initial size. Furthermore it also only is applied when using the same conv type (u/s).

sample that I encontered:
```rust
                let _85816: Idx 4 = %core.conv.u 0 4 _85783;
                let _85817: I64 = %core.conv.u 4 0 _85816;
                let _85818: Idx 4 = %core.conv.u 0 4 _85817;
-> 
                let _85816: Idx 4 = %core.conv.u 0 4 _85783;
```

### DIFF
--- a/src/mim/plug/core/normalizers.cpp
+++ b/src/mim/plug/core/normalizers.cpp
@@ -873,6 +873,14 @@ const Def* normalize_conv(const Def* dst_t, const Def*, const Def* x) {
         return world.lit(d_t, idx_from_signed_mod(*ld, idx_sext(*ls, *l)));
     }
 
+    if (ls && ld)
+        if (auto c1 = Axm::isa(id, x)) {
+            auto x1 = c1->arg();
+            if (auto ls1 = Lit::isa(x1->type()->as<App>()->arg()))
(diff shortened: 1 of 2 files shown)


src/mim/plug/core/normalizers.cpp
```diff
@@ -873,6 +873,14 @@ const Def* normalize_conv(const Def* dst_t, const Def*, const Def* x) {
         return world.lit(d_t, idx_from_signed_mod(*ld, idx_sext(*ls, *l)));
     }
 
+    if (ls && ld)
+        if (auto c1 = Axm::isa(id, x)) {
+            auto x1  = c1->arg();
+            auto ls1 = Lit::isa<u64>(x1->type()->as<App>()->arg());
```
leissa: `u64` is default argument:
```cpp
auto ls1 = Lit::isa(x1->type()->as<App>()->arg());
```
I'm thinking to switch to `u128` as the internal repr, so default arguments are actually more robust.

leissa (approved):
lgtm

PR #438, opened 2026-06-05
Add affine maps to model affine expressions and showcase in map_reduce_aff

fodinabor:
This adds the `%affine.index` / `%affine.map` infrastructure that we discussed a while ago.
It now comes with lowering support as well, no normalizers, yet.
However, as everything is lowered to `Idx`, those normalizers should be able to clean up once lowered.
If we were to implement an analysis of the affine maps, we should probably normalize beforehands, to make that easier (collapse to a single query).

The showcase is a new axiom `%tensor.map_reduce_aff` that could subsumes `%tensor.map_reduce`.
This is a bit more versatile concerning what operations / views can be represented directly (e.g. gather, scatter, tiling, ...).
I can't quite discern what the impact on code gen will be (i.e. is the map evaluation pretty much just fold away, if there's nothing special going on? from what I observed on the test case, yes).

We still generate a few too many `%core.conv.u`s imo, but I believe we can get rid of that with #437.

One thing that I am not sure about, yet, is whether we could actually emit the index calculations in the output's arity? but that would break for things like `(a * b) % c`.

CC @hsqStephenZhang - I believe this should bring us closer to being able to represent a lot of PyTorch operations/views using the `tensor` plugin

### DIFF
--- a/include/mim/plug/affine/phase/lower_index.h
+++ b/include/mim/plug/affine/phase/lower_index.h
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <mim/phase.h>
+
+namespace mim::plug::affine::phase {
+
+/// Lowers the affine index algebra to %core arithmetic.
+///
+/// The opaque %affine.index type is rewritten to the wide `Idx 0` (i64) carrier, and the index operations are computed with
+/// wrap-around (`%core.Mode::none`) arithmetic, so negation/subtraction are correct via two's complement:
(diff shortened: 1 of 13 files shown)


leissa (approved):
lgtm

hsqStephenZhang:
nice, a great part of tensor ops can be represented by the Index operations, both pytorch and tvm use the index mapping to define the semantics of higher-level operations. it's great to have them in mimir!

PR #440, opened 2026-06-08
Bugfix/rewrite order

### LINKED ISSUE #423: [bug] partial eval of recursive lam produces broken specializations
### Reproducer

```sh
./scripts/checkout.sh git@github.com:mimir/mimir.git b62c2b381871ccd2dcf02cf667abaace163ec97f && ./scripts/probe.sh lit/bug/recursive_pe.mim
```

<details>
<summary>relevant Mim code</summary>

```mim
lam make_Ts
    (levels: Nat,
     ns: «levels; Nat»,
     indices: «i: levels; Idx (ns#i)»)@(%core.bit2.and_ 0 (%core.pe.is_closed ns, %core.ncmp.ne (levels, 0)))
  : □
  = let fst = idx 0;
    let done = (%core.ncmp.le (levels, 1));
    let Rec = make_Ts (%core.nat.sub (levels, 1), %tuple.tail ns, %tuple.tail @ levels @ ‹i: levels; Idx (ns#i)› indices);
    let Ts = «i: ns#fst; (★, Rec)#(%core.icmp.e (i, indices#fst))»;
    let Final = «ns#fst; ★»;
    (Ts, Final)#done;
```

</details>

### Observed behavior

Lam recurses with n-1, base case 1 -> recursion should always terminate.
Behavior differs based on whether the current index is closed or not, but both cases use the same recursive application.
The uniform one with open index works without issues, but the closed case produces specializations like the following:

```mim
// copied from mim output
lam make_Ts_10908 (_: Nat, _: «levels_10887; Nat», _: «i_10891: levels_10887; Idx ns_10892#i_10891»): □ =
    «3; make_Ts_10908 (1, 2, ff)»;
```

This is obviously an infinite recursion and therefore incorrect.
I have absolutely no idea why exactly this is happening, but it seems to be caused by the comparison with `i` in the array body. Doing either `[*, *, Ts]` or `«i: 3; (*, Ts)#(i == 2)»` works without issues. 

### Expected behavior

The code should produce a nested type with the structure of stars matching the given indices. For example, the indices `(1_2, 0_3, 2_3)` should produce `[*, [«3; *», *, *]]`.

leissa:
fixes #423 

Order of eval for function calls is impl-defined in C++. This PR fixes this which caused a weird ordering problem in g++ vs clang++ in rewrite.cpp.

### DIFF
--- a/src/mim/rewrite.cpp
+++ b/src/mim/rewrite.cpp
@@ -40,13 +40,17 @@ void Rewriter::reset() {
 }
 
 const Def* Rewriter::map(const Def* old_def, Defs new_defs) {
-    return old2news_.back()[old_def] = world().tuple(new_defs);
+    auto new_tuple                   = world().tuple(new_defs);
+    return old2news_.back()[old_def] = new_tuple;
 }
 const Def* Rewriter::map(Defs old_defs, const Def* new_def) {
-    return old2news_.back()[world().tuple(old_defs)] = new_def;
+    auto old_tuple                     = world().tuple(old_defs);
+    return old2news_.back()[old_tuple] = new_def;
 }
 const Def* Rewriter::map(Defs old_defs, Defs new_defs) {
-    return old2news_.back()[world().tuple(old_defs)] = world().tuple(new_defs);
+    auto old_tuple                     = world().tuple(old_defs);
+    auto new_tuple                     = world().tuple(new_defs);
+    return old2news_.back()[old_tuple] = new_tuple;
 }
 
 const Def* Rewriter::rewrite(const Def* old_def) {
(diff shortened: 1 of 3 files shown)


src/mim/rewrite.cpp
```diff
@@ -94,41 +98,107 @@ DefVec Rewriter::rewrite(Defs ops) {
...
-const Def* Rewriter::rewrite_imm_Meet  (const Meet*   d) { return world().meet  (rewrite(d->ops()));                                                                           }
-const Def* Rewriter::rewrite_imm_Join  (const Join*   d) { return world().join  (rewrite(d->ops()));                                                                           }
+const Def* Rewriter::rewrite_imm_Lit   (const Lit*    d) { return world().lit   (rewrite(d->type()),   d->get());    }
+const Def* Rewriter::rewrite_imm_Match (const Match*  d) { return world().match (rewrite(d->ops()));                 }
+const Def* Rewriter::rewrite_imm_Reform(const Reform* d) { return world().reform(rewrite(d->dom()));                 }
+const Def* Rewriter::rewrite_imm_Sigma (const Sigma*  d) { return world().sigma (rewrite(d->ops()));                 }
+const Def* Rewriter::rewrite_imm_Type  (const Type*   d) { return world().type  (rewrite(d->level()));               }
+const Def* Rewriter::rewrite_imm_UInc  (const UInc*   d) { return world().uinc  (rewrite(d->op()),     d->offset()); }
```
fodinabor, line 106: Formatting is a bit weird now...

fodinabor (approved):
I'm just gonna say ooof... ^^

leissa:
> I'm just gonna say ooof... ^^

I thought this was fixed in C++17 - but no:

> The `postfix-expression` is sequenced before each expression in the expression-list and any default argument. The initialization of a parameter, including every associated value computation and side effect, is **indeterminately** sequenced with respect to that of any other parameter.

PR #443, opened 2026-06-09
Refactor/opt

leissa:
This speeds up `make lit` by more than 3× on my machine in a Debug build. The main reason is that `opt.mim` now constructs the pipeline entirely from strings instead of importing all plugins. Using strings is admittedly somewhat inelegant, but the tradeoff seems worthwhile given the speedup.

Other changes:

* Removed the concept of "magic" plugins. `cli/main.cpp` no longer contains any special-case plugin loading logic. As a consequence, the CLI behavior has changed:

Previously:

```sh
mim foo.mim -O2
mim foo.mim    # -O2 used to be the default
```

Now:

```sh
mim foo.mim -p opt
```

The `opt` plugin is loaded explicitly to enable optimizations.

Similarly, LLVM code generation has changed.

Previously:

```sh
mim foo.mim --output-ll foo.ll
```

Now, the llvm backend lives in its own plugin:

```sh
mim foo.mim -p ll
```

The `ll` plugin is loaded explicitly (which in turn implies loading `opt`). One downside is that specifying the output filename is no longer supported, although we can revisit that in the future if needed.

I also removed the backend plugin hook, since it is no longer necessary, and moved the s-expression emitter into `libmim`.


leissa:
ran out of tokens for copilot - so no copilot review lol

fodinabor (commented):
> The ll plugin is loaded explicitly (which in turn implies loading opt). One downside is that specifying the output filename is no longer supported, although we can revisit that in the future if needed.

That is quite the loss, but yeah, I guess we can revisit that.
@ftschirpke: I guess it would be fine to implement the GPU backend(s) in a similar way? (and name the outpus `<world_name>_nvptx.ll` + `<world_name>_nvptx_embedded.ll` and so on?)

> I also removed the backend plugin hook, since it is no longer necessary, and moved the s-expression emitter into libmim.

What is the reasoning for that? Why not have that work similarly to the ll backend as part of a sexpr plugin?

ftschirpke:
> I guess it would be fine to implement the GPU backend(s) in a similar way? (and name the outpus <world_name>_nvptx.ll + <world_name>_nvptx_embedded.ll and so on?)

From what I can tell, yes. This should work.
Maybe one caveat: This only works as long as we decide that no plugins can depend on eachother, doesn't it?
Of course, that is the current solution. Specifically for the backend we decided to put the whole core LL backend into the header to make it extendable.
This PR would make that even harder to change, I believe, because we then cannot have any magic plugin loads.
But that is the only thing I could think of. This should work fine.

leissa:
> > I also removed the backend plugin hook, since it is no longer necessary, and moved the s-expression emitter into libmim.
> 
> What is the reasoning for that? Why not have that work similarly to the ll backend as part of a sexpr plugin?

I also thought about this. However, there are a couple of issues:
* the ll emission works by placing it as a specific phase into opt.mim. For the sexpr emission we would also have to place it *somewhere* there - and it's unclear to me where exactly.
* I also like to expand the sexpr emission as an additional debug output. Dumping an sexpr would require you to first load the sexpr plugin and do some other things just to see the sexpr output of the program in the debugger (or maybe dumping it somewhere else as debug output).
* The sexpr output currently comes in different flavors: Slotted vs non-slotted, including types or not. And at least at the moment we don't have any way to pass some kind of argument to the plugin.

leissa:
> > I guess it would be fine to implement the GPU backend(s) in a similar way? (and name the outpus <world_name>_nvptx.ll + <world_name>_nvptx_embedded.ll and so on?)
> 
> From what I can tell, yes. This should work. Maybe one caveat: This only works as long as we decide that no plugins can depend on eachother, doesn't it? Of course, that is the current solution. Specifically for the backend we decided to put the whole core LL backend into the header to make it extendable. This PR would make that even harder to change, I believe, because we then cannot have any magic plugin loads. But that is the only thing I could think of. This should work fine.

It's btw no too hard to put code into the cpp file. Say, you have a plugin `bar` that has a function `my_func` that you want to use from `foo`. First you make `bar` a dependency in `foo.mim`:
```
plugin bar;
// ...
```
Then, in `bar` you have somewhere:
```cpp
extern "C" MIM_EXPORT void my_func(World&, int) { /*...*/ }
```
Then, in `foo` you can do:
```cpp
auto my_func = driver.GET_FUN_PTR("bar", my_func);
```
However, note that you are going through `extern "C"` here, so you can't easily extend a C++ class or so (what we need for the GPU stuff). But maybe we can encapsulate bigger chunks of the ll logic like so to move it out of the header.

leissa:
tagging @ashiven here. Maybe he has some thoughts on this subject as well.

ashiven:
I think it doesn't really matter at what point in the compile pipeline the sexpr emitter gets invoked as its output doesn't really depend on any optimizations having been performed. 

I only remember that there was some interaction with the ds2cps phase which required me to run my eqsat phase before the ds2cps phase to internalize rules but that probably isn't relevant here.

Also, wouldn't it be possible to define different phases for the different sexpr flavors similarly to the different backends that I had defined before? But then you would need four different plugins because you can't pass arguments...

Something else I thought about was to have one single sexpr emitter and store the `typed` and `slotted` arguments as compiler flags which would allow us to use one single phase for the plugin. But then you would be mixing plugin arguments into the compiler flags which is also not very pretty..

fodinabor:
> But then you would be mixing plugin arguments into the compiler flags which is also not very pretty..

Actually, I love the fact that in LLVM you can just register `cl::opt` thingys from plugins and then do `clang --pass-plugin <path>.so -mllvm --plugin-specific-opt`. So I'm actually in favor of allowing this - also in light of being able to specify the ll output file name again.

On a different note, not sure when the lit we use actually get's upated... but I did add `%{s:stem}` support upstream as well, so that at some point we can remove the test wrapper from this PR again.
https://github.com/llvm/llvm-project/commit/8f069e7aea7dbf35d87c8a7a700ae6871b831298

leissa:
I'll suggest to merge as is and try to work on sth to allow passing arguments to plugins in a future PR.

PR #444, opened 2026-06-12
Revert "Add axioms to inspect closedness of expressions to refly"

amaebel:
Reverts mimir/mimir#428

After further consideration, those are actually a horrible idea that only leads to pain and suffering. Remove before anyone else tries.

### DIFF
--- a/src/mim/plug/refly/normalizers.cpp
+++ b/src/mim/plug/refly/normalizers.cpp
@@ -100,19 +100,6 @@ const Def* normalize_check(const Def* type, const Def*, const Def* arg) {
     return nullptr;
 }
 
-template<vars id>
-const Def* normalize_vars(const Def* type, const Def* callee, const Def* arg) {
-    auto& world = type->world();
-
-    if constexpr (id == vars::is_closed) return arg->is_closed() ? world.lit_tt() : world.lit_ff();
-
-    if constexpr (id == vars::await_closed) {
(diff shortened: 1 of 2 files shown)


fodinabor:
I was wondering, as an alternative, if it would make sense to add a phase that we can insert wherever that Checks if there are still any refly.check(.is_closed..) Checks around and otherwise executes as if is_closed is false..
Would that make sense to you?

amaebel:
Such a phase would make sense _if_ we want to keep those axioms around, but I strongly advise against doing so. The axioms expose MimIR internals that are not meant to be observable and require reasoning about its exact evaluation order to even attempt to use, which is terrible for a pure functional IR as we have. I am not even sure that there is any way to use them that I would count as correct. I originally added them because they seemed like an attractive hack for a problem I had, where type checking/inference could not deal with the general case. I now fixed this by passing more explicit info to my axiom. 

While the axioms might seem attractive for such trickery, at least the way I wanted to use them was based on oversimplified assumptions about how MimIR works. I don't think they are actually usable in practice and don't want anyone else to be tempted to try.

amaebel:
Also, %core.pe.is_closed is still there and should absolutely be enough to handle any partial eval use cases while not allowing for any cursed shenanigans.

fodinabor:
Yeah, no worries, I'm happy to remove these again - it was just a proposal for something that could replace it with a bit more predictable semantics.

PR #447, opened 2026-06-12
Add admit plugin to allow delaying type checking

amaebel:
Your function is obviously ✨flawless✨ and would absolutely work for any possible input, but the compiler is too simple-minded to tell? With our new trust-me-bro™ plugin you can now just make it accept your code anyway! 

Comes with two amazing axioms:
- `%admit.unify`: Just casts anything to anything else, but actually fails when types are concrete and still can't be unified in contrast to `%core.bitcast`
- `%admit.extract`: You want to extract from something, but the shape of that something is not even computed yet? With `%admit.extract` you can make the compiler wait for that.

Open to adding more similar axioms, but I just needed those two. Possible applications:
- You fold some «n; T» to some other «n; U» by recursively consuming the input and building the output and the compiler cannot even do some basic induction proof? Now you can just `%admit.unify` the result and it _should_ ✨just work✨.
- You compute some type based on a list of indices and then extract with those indices, but your tuple is just some unknown ★ in the general case, causing extract to complain? Use `%admit.extract` to make the compiler just not think about it (yet)!

Obviously this seems a bit hacky, but I think there are tons of applications in MimIR where things are actually type safe but the compiler cannot tell. As we are not a proof assistant but a practical IR, I think that this needlessly limits our expressiveness and having a proper escape hatch is a good idea. It basically allows us to express more at the cost of placing the burden to verify that something type-checks on the user. This does not interfere with type-checking any actual program, as those axioms just normalize themselves away as soon as types are concrete. If your program does not actually type-check, it will still produce a type error.


amaebel:
Can someone explain to me how changing one line of comment in my plugin caused some tensor test to fail on windows? ^^'
This change does not touch any existing code, I am pretty sure the CI failure is unrelated.

src/mim/plug/admit/normalizers.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+template<extract> const Def* normalize_extract(const Def* type, const Def* callee, const Def* arg) {
+    auto& world = type->world();
+    auto tup    = callee->as<App>()->arg();
+    auto t      = tup->type()->zonk();
+    if (!t->is_closed()) return nullptr;
+
+    return world.extract(tup, arg);
+}
+
```
leissa, line 23: I think zonking is unnecessary here as World:app has already zonked everything before running the normalizer.

leissa (commented):
otherwise lgtm

leissa:
> Can someone explain to me how changing one line of comment in my plugin caused some tensor test to fail on windows? ^^' This change does not touch any existing code, I am pretty sure the CI failure is unrelated.

~~I think it's only a timeout. This is why Windows release isn't a problem.~~

Asked my AI overlord:

> So this is a flaky Windows-only environmental failure (intermittent DLL load / process spawn returning 127), not a timeout and not caused by the change. The PR author's instinct that the failure is unrelated is correct — but the mechanism is a spurious 127, not a timeout.

PR #457, opened 2026-06-23
Refactor/sexpr

ashiven:
Adjusts the output of the regular `sexpr` backend to resemble the output of the `slotted-sexpr` backend more closely.

- Variable projections are always explicitly let-bound
- Types are only emitted in the annotated version
- `extern` is emitted as part of a root-level binder instead of the lambda


leissa (approved):
lgtm 👍

PR #458, opened 2026-06-26
Respect scalarize threshold for nested arrays

fodinabor:
This change makes sure that the flatten utilities don't just check the current rank's size but the full nested size which it would recursively flatten to.
I.e. `lam f (a : <4;<4;<4; Nat>>>) : ...` was scalarized to 64 scalar values independent of the scalarize threshold.. which is maybe not so great :)

This surfaced a bug in SymExprOpt where `con f (a : <32; Nat>, r: Cn [<32;Nat]) = r a` get's `r` propagated. The `num_vars` of the new lam is then 32, but the actual args that get rewritten are just 2 and one of them is skipped -> illegal memory accesses to the empty 31 arguments.


src/mim/phase/sym_expr_opt.cpp
```diff
@@ -168,12 +168,11 @@ const Def* SymExprOpt::rewrite_imm_App(const App* old_app) {
...
-            for (size_t i = 0, j = 0; i != num_old; ++i) {
+            auto new_args  = DefVec{};
+            for (size_t i = 0; i != num_old; ++i) {
                 auto old_var = old_lam->var(num_old, i);
                 auto abstr   = lattice(old_var);
-                if (keep(old_var, abstr)) new_args[j++] = rewrite(old_app->targ(i));
+                if (keep(old_var, abstr)) new_args.emplace_back(rewrite(old_app->targ(i)));
             }
 
             return map(old_app, new_world().app(new_lam, new_args));
```
leissa, line 178: can you revert this part? This will give us hairy conflicts in #407 😱

src/mim/tuple.cpp
```diff
@@ -103,7 +119,7 @@ std::string tuple2str(const Def* def) {
 // TODO flatten/unflatten needs to be rewritten
 
 size_t flatten(DefVec& ops, const Def* def, bool flatten_muts) {
-    if (auto a = Lit::isa(def->arity()); a && *a != 1 && should_flatten(def) && flatten_muts == mut_val_or_typ(def)) {
+    if (auto a = Lit::isa(def->arity()); a && *a != 1 && should_flatten(def) && flatten_muts == mut_val_or_type(def)) {
         auto n = 0;
```
leissa, line 123: This whole code is quite old and messy and could definitely require a face lift. But let's keep this for another day.

leissa (approved):

src/mim/phase/sym_expr_opt.cpp
fodinabor: well, the current version is just broken ... so not sure if it'd be a good idea to keep it.. 
The file get's deleted in that PR.. so it should be a pretty easy merge

PR #459, opened 2026-06-26
feat(tensor): batch mm

hsqStephenZhang:


fodinabor (commented):
Apart from the naming of the operation, it looks good: I think, I'd prefer to make clear what it does, i.e. `batched_product_2d` or `bmm` ... it's not really a 3d product, just a batch of 2d products (at least that's my understanding).

hsqStephenZhang:
done.

PR #460, opened 2026-06-26
GPU Axioms v2

ftschirpke:
This PR changes existing axioms for GPU computing and introduces new ones. These changes are the product of the development of a NVIDIA GPU backend, where I noticed that some we require additional features and some more expressiveness.

Namely, this PR...
- ... merges asynchronous and blocking operations into one and uses tags `.block` and `.async`
- ... adds explicit operations for GPU context initialization and deinitialization
- ... adds support for symbols (which are global variables in CUDA) while preserving the dataflow of the functional IR
- ... merges the two versions of the kernel launch operation `%gpu.launch` and adds new arguments
- ... adjusts all existing test cases to the changes above and adds new test cases as well


ftschirpke:
Both, the Windows and the MacOS build, have a problem with the async tag when creating python binding, it seems like:
```
E     File "/Users/runner/work/mimir/mimir/build/mim_py_stage/main/src/mim/_plugins/gpu.py", line 36
E       		async = 0x2115800000001801
E       		      ^
E   SyntaxError: invalid syntax
```
I am not sure how to fix this, as I am not proficient in Python bindings. Help would be appreciated :)

fodinabor:
Maybe a somewhat design related discussion: The symbol pointer stuff, for example, feels as if it might be somewhat CUDA (API) specific?   
That raises the question to me, how we should handle inconsistencies between what different (prospective) backends/runtimes might support:
- delegate things that are not general GPGPU concepts (like launch, stream, id, ..) to backend-specific plugins?
- include in the gpu plugin, but document what's (not) supported?
- include things in the generic plugin only, if it is natively supported by backends or there is a straight-forward emulation

ftschirpke:
> Maybe a somewhat design related discussion: The symbol pointer stuff, for example, feels as if it might be somewhat CUDA (API) specific?

Very fair, I did not look into how other vendors do this, but a similar approach to implement global variables might still be helpful, just not like this?

> That raises the question to me, how we should handle inconsistencies between what different (prospective) backends/runtimes might support:
> 
> * delegate things that are not general GPGPU concepts (like launch, stream, id, ..) to backend-specific plugins?

I really thought there might be a way to provide vendor-independent abstractions? Of course, I am currently very NVIDIA-focused in the designs.
But simply swapping out the backend would be very nice, wouldn't it?

> * delegate things that are not general GPGPU concepts (like launch, stream, id, ..) to backend-specific plugins?

On the other hand, the backends have to lower the concepts differently anyway, so I also see a lot of advantages in just building backend-specific plugins.

fodinabor:
Maybe a part of my message was easily mis-interpretable :D
> delegate things that are not general GPGPU concepts (like launch, stream, id, ..) to backend-specific plugins?

I meant parts except the launch/stream/id stuff :D

So what's a bit less portable is stuff like those symptrs, which are somewhat specific, or maybe intrinsics for warp operations and so on - although one can usually find common abstractions.

Regarding the need of the symbol pointers: I don't necessarily see the big need for it.. I never really used them.. 🤷🏼
The idea behind some of the constant / symptr things is that this info could actually be baked into the final binary, which can give some benefits.. but I'd rather go with a full specializing JIT at that point.. :)

ftschirpke:
Okay, I think then are mostly on the same page.
I think, until we find these common abstractions, we should definitely move warp intrinsics etc. to backend-specific plugins. We can unify them later, right?
I was just wondering about symbols specifically, because I thought they are deeply intertwined with the kernel launch, which they are totally not. I can remove SymPtr from the `%gpu.launch` type. I have a change in mind, but would only start implementing this tomorrow.
Is there anything else besides symbols (the SymPtr) you would want to address?
I definitely would be interested in what I need to do to fix the compiler error on windows and mac, maybe @leissa knows better?

fodinabor:
Apart from that, I'm happy with the PR :)

@JoeJoeSch might also know more w.r.t. the Python things

JoeJoeSch:
yeah this looks like an issue with the naming, since the there will be an axiom or axiom member generated just called "async" this is an issue since that is a python keyword, would be great if you could add a pre- or suffix to that

ftschirpke:
I see, that makes sense. Thank you.
Weird that this was not a problem for the build in linux.
Well, does anyone have a good idea for a new name of this tag?
With the full word, something like %gpu.alloc.asynchronous is pretty long.
Maybe just add the h for asynch? Or rather asyn?

JoeJoeSch:
yeah that is indeed weird, Iam going to look into that

ftschirpke:
I now have a working implementation with symbols separated from the initialization and launch operations in my fork and thus removed the symbol pointers here.
Also, renaming the tags from `async` to `asyn` indeed solved the build problems.

fodinabor (approved):
LGTM :)

leissa:
guess we can merge :)

PR #461, opened 2026-06-26
feat(tensor): split tensor.get/set lowering into a separate lower_get_set

hsqStephenZhang:
## Background

1. TVM has **high level** IR, where the tensor access is not lowered to index expr calculation and access, making the IR easy to understand. Here is a small demo:

```python
@I.ir_module
class Module:
    @T.prim_func(s_tir=True)
    def main(A: T.Buffer((128,), "float32"), B: T.Buffer((128,), "float32"), C: T.Buffer((128,), "float32")):
        T.func_attr({"tirx.noalias": True})
        # with T.sblock("root"):
        for i in range(128):
            with T.sblock("C"):
                v_i = T.axis.spatial(128, i)
                T.reads(A[v_i], B[v_i])
                T.writes(C[v_i])
                C[v_i] = A[v_i] + B[v_i]
```

2. The current Tensor plugin eagerly lowers get/set operations, generating many extract/insert nodes. Many of these are temporary nodes with names like `_76786` `_76834#tt#tt`, which makes the generated IR hard to read.

## Solution

1. make `%tensor.get/set` a separate phase than the `LowerMapReduce` so we could customize the `_compile` Phase to keep the `%tensor.get/set` in output IR.
2. moreover, `nested_extract` and `nested_insert` could be implemented via the `op_get/set`, reducing redundancy. 

## Test

1. test via the following case, you could see that `%tensor.get/set` is kept. 

```mimir
plugin core;
plugin tensor;
plugin affine;
plugin mem;
plugin direct;
import compile;

fun copy (x: I32, y: «1; I32»)@tt = return (y#0_1);
lam id2 (i,j: %affine.index): «2; %affine.index» = (i, j);
lam tr  (i,j: %affine.index): «2; %affine.index» = (j, i);

fun extern transpose (input: «2, 3; I32»): «3, 2; I32» =
    return (%tensor.map_reduce_aff @1 @(I32, 2, 0) ((3,2), (2, 3))
                @(‹1; I32›, ‹1; 2›, ‹1; (2, 3)›) (copy, ⊥:I32) tr id2 input);

// don't lower %tensor.get/set
lam extern _compile(): %compile.Phase =
    %compile.phases ff (
        %compile.named_phase "%tensor.lower_tensor",
        %compile.named_phase "%tensor.lower_map_reduce",
    );
```

2. the integration lit test fails since the `origin/feature/tensor-ops` had some regressions that is not fixed.

### DIFF
--- a/src/mim/plug/tensor/phase/lower_map_reduce.cpp
+++ b/src/mim/plug/tensor/phase/lower_map_reduce.cpp
@@ -15,87 +15,74 @@
 
 namespace mim::plug::tensor::phase {
 
-const Def* LowerMapReduce::lower_get(const App* app) {
-    auto& w  = new_world();
-    auto c   = rewrite(app->callee());
-    auto arg = rewrite(app->arg());
+const Def* LowerMapReduce::rec_broadcast(const Def* s_in, const Def* s_out, const Def* input, u64 r, u64 i) {
+    auto& w = new_world();
+    // Base case: all dimensions have been processed; `input` is the final scalar.
+    if (i == r) return input;
 
-    auto [arr, index] = arg->projs<2>();
-    auto callee       = c->as<App>();
-    auto [T, r, s]    = callee->args<3>();
+    auto s_in_ri = s_in->proj(r, i), s_out_ri = s_out->proj(r, i);
+    DLOG("rec_broadcast");
+    DLOG("    r = {}", r);
+    DLOG("    i = {}", i);
+    DLOG("    s_in_ri = {} : {}", s_in_ri, s_in_ri->type());
+    DLOG("    s_out_ri = {} : {}", s_out_ri, s_out_ri->type());
+    DLOG("    input = {} : {}", input, input->type());
 
-    DLOG("lower_get");
-    DLOG("    arr = {} : {}", arr, arr->type());
-    if (auto arr_seq = arr->type()->isa<Seq>()) DLOG("    arr shape = {}", arr_seq->arity());
-    DLOG("    index = {} : {}", index, index->type());
-    DLOG("    T = {} : {}", T, T->type());
-    DLOG("    r = {} : {}", r, r->type());
-    DLOG("    s = {} : {}", s, s->type());
+    if (s_in_ri == s_out_ri) {
+        if (auto s_in_lit = Lit::isa<u64>(s_in_ri)) {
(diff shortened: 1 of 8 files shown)


hsqStephenZhang:
And I found the explosion problem when lowering conv could be solved by this approach(keeping the %tensor.get/set without lowering into array extract/insert)

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -176,33 +176,28 @@ static std::pair<Lam*, const Def*> counting_for(const Def* bound, const Def* acc
...
-static DefVec live_coords(const Def* coords, const Def* shape, u64 r) {
-    DefVec idx;
-    for (u64 k = 0; k < r; ++k) {
-        if (auto d = Lit::isa<u64>(shape->proj(r, k)); d && *d == 1) continue;
-        idx.push_back(coords->proj(r, k));
+
+static const Def* get_element_type(const Def* type, u64 r) {
+    auto cur = type;
+    for (u64 i = 0; i < r; ++i) {
+        if (cur->node() == Node::Arr || cur->node() == Node::Pack) {
```
fodinabor: Please read some code and the developer docs to see how we do checks on types and axioms... we pretty much never do explicit `node() == Node::XX` checks.

```suggestion
        if(auto seq = cur->isa<Seq>()) cur = seq->body();
```

fodinabor (commented):
Thanks!
two points:
1. we generally split each phase into its own set of header/source files.
2. is it actually solving the explosion - or just delaying it?

hsqStephenZhang:
the code has been re-organized.

regarding the second point, current approach only delays the explosion. If we treat tensor as multi-dim array and generate `Extract/Insert` for each operation, then there surely will be myriads of nodes and the explosion is unavoidable. We might need a bufferization plugin to resolve this issue once for all.

hsqStephenZhang:
since `%core.nat.div` is [supported](https://github.com/mimir/mimir/commit/e6958c455b2223edfadce36126cf3be2b9c10fad), i'd like to fix the output tensor shape problem(output tensor's shape should be calculated instead of passed) of `%tensor.conv` now. 

but it would be painful to have the full lowering phase fully expand the `%tensor.get/set`, which could lead to compile time explosion, could we get this merged first? cc @fodinabor

PR #463, opened 2026-06-28
Bugfix/curried rec pe

leissa:


leissa:
superseded by #464.

We might later revisit the Def::zonk_ magic in this branch though.

PR #466, opened 2026-06-30
GPU Phases

ftschirpke:
This PR implements phases for the GPU plugin that perform safety checks that we were not able to encode in the type system and perform generic preparation for GPU backends.

Specifically, we implement the phases:
- `MemChecks` - checks and validates that no `%mem.M` is passed across device binaries (cpu <-> gpu) or a device `%mem.M`used as an argument or return value of the host's main function
- `RemoveDoubleSyncs` - removes consecutive synchronization operations, which could be introduced by the `%gpu.synced_scope` abstraction
- `SplitOffKernels` - splits off the device code (kernel function and everything they depend on) from the host code, creating two new worlds
- `SplitApply` - uses the `SplitOffKernels` phase to split the code and then applies host-specific phases to the host code and device-specific phases to the device code

As only `SplitOffKernels` is strictly necessary for further development of GPU backends and the other phases are mostly niceties, I would like to have a discussion about what makes sense to merge and what is unnecessary or could/should be solved in a different way.

### DIFF
--- a/include/mim/plug/gpu/phase/mem_checks.h
+++ b/include/mim/plug/gpu/phase/mem_checks.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include "mim/phase.h"
+
+#include "mim/plug/gpu/gpu.h"
+
+namespace mim::plug::gpu::phase {
+
+class MemChecks : public Analysis {
+public:
+    using Super = Analysis;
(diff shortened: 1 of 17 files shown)


leissa (approved):
lgtm 👍

PR #484, opened 2026-07-13
Add support for encoding decimal float literals with correct bit width.

fodinabor:
I ended up running into a bunch of nan errors - which seem to stem from encoding 32-bit floats as a bit representation of a double.
The normalizers (and maybe also backend) seem to be confused by this, and thus did invalid math -> nan...


leissa:
it's kind of hacky, because now mimir-base knows of the math plugin - but still lightyears better than what we have now XD

leissa (approved):
👍

leissa:
just tagging @HarithF here, as we talked about this the other day

PR #485, opened 2026-07-14
NVPTX backend

ftschirpke:
This PR implements a backend that extends the LLVM IR emitter from the `ll` plugin with NVPTX intrinsics to target NVIDIA GPUs.

Namely, this PR:
- adjusts the `ll` plugin emitter to work for multiple address spaces
- adds the `ll_nvptx` plugin containing the backend phase
- adds the `nvptx` plugin for NVPTX we want to expose in the IR (none currently, but more in the future)
- adds and configures tests for the `ll_nvptx` backend that require a Linux system and the CUDA toolkit

What is left to do:
- [x] fix the interaction between externalizing GPU kernels and the `%compile.eta_conv` phase that causes two functions of the same name to be externalized (which is not allowed) when a kernel is launched multiple times
- [x] (maybe) integrate tests into the CI (GPU runners are pricey)

I will also open issues for some more open problems that we might need to address in the long term.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -15,6 +15,30 @@ Mim looks for plugins in this order:
 4. `path/to/mim.exe/../../lib/mim`
 5. `CMAKE_INSTALL_PREFIX/lib/mim`
 
+## Passing Arguments to Plugins & Phases {#clipluginargs}
+
+Plugins - and in particular backends - often need to be configured from the command line.
+For example, a backend that invokes an external tool may want to forward optimization levels, a target triple for cross-compilation, or library paths.
+Use `-X` / `--plugin-arg` for this:
+
+```
+mim foo.mim -p ll -X ll:o=out.ll -X compile:aggr=on
+```
+
+The syntax is `-X <plugin>:<arg>`:
+
+- The option is repeatable; each occurrence contributes one argument.
+- Only the *first* `:` separates `<plugin>` from `<arg>`, so `<arg>` may itself contain `:` or `=` (e.g. Windows paths or `key=value` pairs).
+- Arguments are keyed by plugin name and collected on the [`mim::Driver`](@ref mim::Driver).
+  A [`mim::Phase`](@ref mim::Phase) reads the arguments addressed to its own plugin via [`mim::Phase::args`](@ref mim::Phase::args); the interpretation of each `<arg>` is up to the plugin.
+
+### Known Arguments
+
+| Plugin              | Argument      | Effect                                                                                                                   |
+|---------------------|---------------|--------------------------------------------------------------------------------------------------------------------------|
+| [compile](@ref compile) | `aggr=<bool>` | Value of `%%compile.aggr`; toggles fixed-point iteration of the `opt` pipeline's `optimize` stage (default off). `<bool>` is `on`/`tt`/`true` or `off`/`ff`/`false`; a bare `aggr` means `on`. See @ref compile_cli_args. |
+| [ll](@ref ll)           | `o=<file>`    | Write the LLVM IR to `<file>` instead of the default `<world>.ll`/`a.ll`. See @ref ll_cli_args.                          |
+
 ## Debugging Features {#clidebug}
 
 - The breakpoint-oriented flags below are developer options that are only available when MimIR is built with `MIM_ENABLE_CHECKS`.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -38,6 +38,14 @@ The syntax is `-X <plugin>:<arg>`:
 |---------------------|---------------|--------------------------------------------------------------------------------------------------------------------------|
 | [compile](@ref compile) | `aggr=<bool>` | Value of `%%compile.aggr`; toggles fixed-point iteration of the `opt` pipeline's `optimize` stage (default off). `<bool>` is `on`/`tt`/`true` or `off`/`ff`/`false`; a bare `aggr` means `on`. See @ref compile_cli_args. |
 | [ll](@ref ll)           | `o=<file>`    | Write the LLVM IR to `<file>` instead of the default `<world>.ll`/`a.ll`. See @ref ll_cli_args.                          |
+| [ll_nvptx](@ref ll_nvptx) | `o=<file>`         | Write the (host) LLVM IR to `<file>` instead of the default `<world>.ll`/`a.ll`. See @ref ll_nvptx_cli_args.                             |
+| [ll_nvptx](@ref ll_nvptx) | `o-dev=<file>`     | Write the device LLVM IR to `<file>` instead of the default `<world>_dev.ll`/`a_dev.ll`. See @ref ll_nvptx_cli_args.                     |
+| [ll_nvptx](@ref ll_nvptx) | `no-embed`         | Don't embed compiled device binary in host LLVM IR, which is the default behavior. See @ref ll_nvptx_cli_args.       |
+| [ll_nvptx](@ref ll_nvptx) | `no-ptx-embed`     | If embedding device binary, don't embed PTX image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.            |
+| [ll_nvptx](@ref ll_nvptx) | `no-cubin-embed`   | If embedding device binary, don't embed CUBIN image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.          |
+| [ll_nvptx](@ref ll_nvptx) | `sm=<SM>`          | If embedding device binary, compile the device binary for compute capability sm_<SM>. See @ref ll_nvptx_cli_args.                  |
+| [ll_nvptx](@ref ll_nvptx) | `libdevice=<path>` | If embedding device binary and if linking libdevice, link against the libdevice NVVM library at <path> instead of trying to find it via CUDA paths. See @ref ll_nvptx_cli_args. |
+| [ll_nvptx](@ref ll_nvptx) | `X<tool>=<args>`   | If embedding device binary invoke <tool> with <args>. See @ref ll_nvptx_cli_args for more details.     |
 
 ## Debugging Features {#clidebug}
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -43,9 +43,9 @@ The syntax is `-X <plugin>:<arg>`:
 | [ll_nvptx](@ref ll_nvptx) | `no-embed`         | Don't embed compiled device binary in host LLVM IR, which is the default behavior. See @ref ll_nvptx_cli_args.       |
 | [ll_nvptx](@ref ll_nvptx) | `no-ptx-embed`     | If embedding device binary, don't embed PTX image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.            |
 | [ll_nvptx](@ref ll_nvptx) | `no-cubin-embed`   | If embedding device binary, don't embed CUBIN image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.          |
-| [ll_nvptx](@ref ll_nvptx) | `sm=<SM>`          | If embedding device binary, compile the device binary for compute capability sm_<SM>. See @ref ll_nvptx_cli_args.                  |
-| [ll_nvptx](@ref ll_nvptx) | `libdevice=<path>` | If embedding device binary and if linking libdevice, link against the libdevice NVVM library at <path> instead of trying to find it via CUDA paths. See @ref ll_nvptx_cli_args. |
-| [ll_nvptx](@ref ll_nvptx) | `X<tool>=<args>`   | If embedding device binary invoke <tool> with <args>. See @ref ll_nvptx_cli_args for more details.     |
+| [ll_nvptx](@ref ll_nvptx) | `sm=<SM>`          | If embedding device binary, compile the device binary for compute capability sm_`<SM>`. See @ref ll_nvptx_cli_args.                  |
+| [ll_nvptx](@ref ll_nvptx) | `libdevice=<path>` | If embedding device binary and if linking libdevice, link against the libdevice NVVM library at `<path>` instead of trying to find it via CUDA paths. See @ref ll_nvptx_cli_args. |
+| [ll_nvptx](@ref ll_nvptx) | `X<tool>=<args>`   | If embedding device binary invoke `<tool>` with `<args>`. See @ref ll_nvptx_cli_args for more details.     |
 
 ## Debugging Features {#clidebug}
 
--- a/include/mim/plug/ll/ll.h
+++ b/include/mim/plug/ll/ll.h
@@ -228,8 +228,7 @@ inline std::string Emitter::convert(const Def* type, bool simd) {
         }
     } else if (auto ptr = Axm::isa<mem::Ptr>(type)) {
         auto [pointee, addr_space] = ptr->args<2>();
-        // TODO addr_space
-        std::print(s, "{}*", convert(pointee, false));
+        std::print(s, "{} addrspace({})*", convert(pointee, false), addr_space);
     } else if (auto arr = type->isa<Arr>()) {
         if (auto se = is_simd(arr); se && simd) {
             auto [size, elem] = *se;
@@ -397,7 +396,9 @@ inline void Emitter::emit_epilogue(Lam* lam) {
 
         switch (values.size()) {
             case 0: return bb.tail("ret void");
-            case 1: return bb.tail("ret {} {}", convert(types[0]), values[0]);
+            case 1:
+                return Axm::isa<mem::M>(types[0]) ? bb.tail("ret void")
+                                                  : bb.tail("ret {} {}", convert(types[0]), values[0]);
             default: {
                 std::string type;
                 std::string prev;
@@ -425,6 +426,7 @@ inline void Emitter::emit_epilogue(Lam* lam) {
                     prev = "undef";
                     type = convert(world().sigma(types));
                     for (size_t i = 0, n = values.size(); i != n; ++i) {
+                        if (auto mem = Axm::isa<mem::M>(types[i])) continue;
                         auto v_elem = values[i];
                         auto t_elem = convert(types[i]);
                         auto namei  = "%ret_val." + std::to_string(i);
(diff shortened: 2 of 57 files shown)


src/mim/plug/ll_nvptx/ll_nvptx.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto stage         = (*stage_funcptr)(world);
+    auto phase         = stage->isa<RWPhase>();
+    if (!phase) error("Found unexpected gpu::split_apply stage");
+    return std::make_pair(std::move(stage), phase);
+}
+
+constexpr auto default_compute_cap = "75";
+
+static std::string get_compute_capability() {
+    auto out      = sys::exec("nvidia-smi --query-gpu=compute_cap");
```
hsqStephenZhang: do we need `--format=csv,noheader`?

lit/nvptx/array_kernel.mim
```diff
@@ -0,0 +1,53 @@
...
+    let (m0, m1) = %gpu.copy_to_host.block (m0, m1, d_ptr, arr);
+    let m1 = %mem.free (m1, d_ptr);
+
+    let m0 = %gpu.deinit (m0, m1, m4);
+
+    let idx = %core.bitcast (Idx 4) (%core.wrap.sub 0 (argc, 1I32));
+    let val_arr = %mem.lea (arr, idx);
+    let (m0, val) = %mem.load (m0, val_arr);
+    let m0 = %mem.free (m0, arr);
+
```
hsqStephenZhang, line 52: can we have helpers like `assert_close` to check the result of f32/f64?

include/mim/plug/ll/ll.h
```diff
@@ -43,7 +43,7 @@ namespace math = mim::plug::math;
 namespace mem  = mim::plug::mem;
 namespace vec  = mim::plug::vec;
 
-namespace {
+namespace internals {
```
fodinabor: we typically are using `detail` for these things.

lit/lit.cfg.py
```diff
@@ -23,3 +23,10 @@
 # lit.site.cfg.py. Tests emitting %math.F16 literals gate on this via `// REQUIRES: fp16`.
 if getattr(config, 'fp16', False):
     config.available_features.add("fp16")
+# operating system is detected using sys.platform (see https://docs.python.org/3/library/sys.html#sys.platform)
+# Tests that only work on e.g. Linux operating systems specify this via `// REQUIRES: system-linux`.
+config.available_features.add(f"system-{sys.platform}")
+# CUDA support (needs CUDA toolkit) is detected once by CMake and passed in via
+# lit.site.cfg.py. Tests gate on this via `// REQUIRES: cuda`.
+if getattr(config, 'cuda_found', False):
+    config.available_features.add('cuda')
```
fodinabor: hmm :)
I guess this will get annoying on my laptop, as I don't have an Nvidia GPU but do have CUDA installed to be able to build things against CUDA...

We could do sth fun like the following in CMake - or actually also here in the lit cfg. 🤔 
```cmake
find_program(NVIDIA_SMI nvidia-smi)
if(NVIDIA_SMI)
  execute_process(COMMAND ${NVIDIA_SMI} -L
    OUTPUT_VARIABLE _smi_out RESULT_VARIABLE _smi_rc
    OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
  if(_smi_rc EQUAL 0 AND _smi_out MATCHES "GPU [0-9]+:")
    set(CUDA_CAN_RUN_ON_DEVICE ON)
  endif()
endif()
```

src/mim/plug/ll_nvptx/ll_nvptx.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+#include "mim/plug/ll_nvptx/ll_nvptx.h"
+
+using namespace std::string_literals;
+
+namespace mim::plug::ll_nvptx {
+
+namespace {
+
+static auto get_setup_phase(World& world) {
```
fodinabor: `namespace {` already makes these functions TU internal. no need for `static`.

src/mim/plug/ll_nvptx/ll_nvptx.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto flags         = Annex::base<gpu::split_apply>();
+    auto stage_funcptr = world.driver().phase(flags);
+    auto stage         = (*stage_funcptr)(world);
+    auto phase         = stage->isa<RWPhase>();
+    if (!phase) error("Found unexpected gpu::split_apply stage");
+    return std::make_pair(std::move(stage), phase);
```
fodinabor: Shouldn't this do as well?
```c++
Phase::create(world.driver().phases(), world.annex<gpu::split_apply>());
```

src/mim/plug/ll_nvptx/ll_nvptx.cpp
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        auto compute_cap = get_compute_capability();
+
+        bool embed_device_code;
+#ifdef _WIN32
+        embed_device_code = false;
+#elif defined(__APPLE__)
+        embed_device_code = false;
+#elif defined(__linux__)
```
fodinabor: Invert -> `#ifdef __linux__` -> true `#else` -> false.

fodinabor (commented):
Looks quite clean already 👍🏼
Just a few style things and so on.

lit/nvptx/array_kernel.mim
```diff
@@ -0,0 +1,53 @@
...
+            return: Cn [%gpu.GlobalM, %gpu.SharedM, %gpu.ConstM, %gpu.LocalM]) =
+    let el = %mem.lea (data, item_id);
+    let val = %core.bitcast I32 item_id;
+    let val = %core.wrap.add 0 (val, 42I32);
+    let m1 = %mem.store (m1, el, val);
+    return (m1, m3, m4, m5);
+
+fun extern main (m0: %mem.M 0, argc: I32, argv: %mem.Ptr («⊤:Nat; %mem.Ptr («⊤:Nat; I8», 0)», 0))
+              : [%mem.M 0, I32] =
+    let (m0, arr) = %mem.alloc («4; I32», 0) m0;
```
hsqStephenZhang, line 25: maybe have a HOF of wrapper for `alloc + launch kernel + dealloc` to make it a general gpu coding style?

fodinabor: > HOF

?

hsqStephenZhang: higher-order function (this could partial address the api/axiom misuse problem)

src/mim/plug/ll_nvptx/ll_nvptx.cpp
ftschirpke: While this on its own does not change anything as the output is still split over multiple lines, it inspired me to make this parsing step more readable. I hope this resolves your comment?

lit/nvptx/array_kernel.mim
ftschirpke: I am unsure what you are referring to specific to this test case or PR.
I understand that such an assertion would be nice, but here we are dealing with integer values and check the tests using the return values of the main functions (which need to be integers).
I believe this is not a concern for the NVPTX backend either way.

lit/nvptx/array_kernel.mim
ftschirpke: From my point of view, the goal was to provide low-level GPU functionality i.e. we did not want to abstract such control away completely.
And we can certainly design abstractions that hide this, but this would most definitely be out of scope for this backend.
Instead, we should discuss this with the design of the GPU plugin (see #460) and make additions/adjustments there.

lit/lit.cfg.py
ftschirpke: Nice recommendation, thank you. I pretty much fully copied your code and tested it on my laptop. Everything works (on my systems :wink:)

ftschirpke:
I rebased the branch to align with the changes on the master branch, especially with the changes to `%mem.slot` and `%mem.mslot`.
In my view, this leaves only the ~two~ open problem~s~ stated right in the first PR message:

> What is left to do:
> - [x] fix the interaction between externalizing GPU kernels and the %compile.eta_conv phase that causes two functions of the same name to be externalized (which is not allowed) when a kernel is launched multiple times
> - [x] (maybe) integrate tests into the CI (GPU runners are pricey)

~I was hoping, @leissa, you would know what to do about the `%compile.eta_conv` phase to make the new names unique?~
And on the regards of integrating GPU tests into the CI, I still have no good idea besides just not doing it. Does anyone have any other ideas?

leissa:
We can at least run FileCheck based tests.

ftschirpke:
It looks like I finally worked out all the kinks.
As the commit messages show, I applied a lot of small fixes and added tests that check the emitted LLVM IR via FileCheck.
Still, for the PR to be merged and in case I (re)introduced any problems, I would greatly appreciate another set of reviews @leissa @fodinabor @hsqStephenZhang.
Thank you.

leissa:
powered by claude:

# Code Review — `feature/nvptx-backend` vs `mimir/master`

Reviewed the full diff (NVPTX/CUDA backend: new `nvptx` + `ll_nvptx` plugins, `ll::Emitter` addr-space changes, gpu phase tweaks, lit tests). High-effort recall pass. Findings ranked most-severe first.

---

### 1. `find_libdevice`: uncaught `filesystem_error` when `nvcc` is absent — crashes instead of falling back
`src/mim/plug/ll_nvptx/ll_nvptx.cpp:84-85`

```cpp
auto nvcc_which = sys::find_cmd("nvcc");
if (nvcc_which.find("not found") == std::string::npos) {
    auto nvcc_path = std::filesystem::canonical(nvcc_which);
```

`sys::find_cmd` runs `which nvcc` and captures **stdout only**; when `nvcc` is missing it returns an **empty string**, never a string containing `"not found"`. So `""`.find(...) == npos → the branch is entered → `std::filesystem::canonical("")` throws a `std::filesystem::filesystem_error`.

That exception type is **not** `CmdNotFound`, so it sails past the `catch (const CmdNotFound&)` fallback in `Emit::start` and aborts the compiler. **Failure scenario:** a machine with a GPU (so `get_compute_capability` succeeds) but no CUDA toolkit, compiling a libdevice-using kernel → hard crash instead of the intended "Falling back to not embedding device code." Every other call site in this file checks `std::filesystem::exists(cmd)`; this one should too.

---

### 2. Device sreg intrinsic `tid.x` is called but never declared → invalid device IR
`src/mim/plug/ll_nvptx/phase/ll_nvptx.cpp:522` (and the call at ~526)

```cpp
declare("i32 @llvm.nvvm.read.ptx.sreg.ctaid.x()");   // hardcoded
...
bb.assign(name, "call i32 @llvm.nvvm.read.ptx.sreg.{}()", sreg);  // uses {sreg}
```

`register_sreg_idx` is called for both `group_id` (`sreg="ctaid.x"`) and `item_id` (`sreg="tid.x"`), but the `declare` string is hardcoded to `ctaid.x` in both invocations. `declare` dedups, so `@llvm.nvvm.read.ptx.sreg.ctaid.x` gets declared (twice) while `@llvm.nvvm.read.ptx.sreg.tid.x` is **called with no declaration**. **Failure scenario:** the emitted `*_dev.ll` references an undeclared global; `clang`/`llc` fails to parse it ("use of undefined value"). This is masked in CI because the non-`cuda` lit runs only FileCheck the text — it only bites on a real CUDA machine (`REQUIRES: cuda`). Should be `declare("i32 @llvm.nvvm.read.ptx.sreg.{}()", sreg);`.

---

### 3. `DeviceEmitter::convert` changes the overridden default arg `simd` from `true`→`false`
`src/mim/plug/ll_nvptx/phase/ll_nvptx.cpp:74`

```cpp
std::string convert(const Def* def, bool simd = false) override { return Super::convert(def, simd); }
```

Base is `virtual std::string convert(const Def*, bool simd = true)`. Default arguments bind **statically** to the caller's declared type. All the `convert(x)` calls inside `ll::Emitter` base methods (`emit_bb`, `emit_epilogue`, `prepare`, `convert_ret_pi`, …) still substitute `true`, then virtually dispatch into `DeviceEmitter::convert(x, /*simd=*/true)`. So the intent ("no SIMD on the NVPTX device") only holds for calls written inside `DeviceEmitter`; every base-class code path still requests `simd=true`. **Failure scenario:** a device sigma/array that `is_simd` reaches `convert` via base code and is emitted as an LLVM vector type on a target where that wasn't intended — inconsistent codegen and a classic C++ footgun. Force the value at the base-call boundary instead of redefining the default.

---

### 4. `get_compute_capability`: dot-position `assert` is compiled out in release → garbage compute cap
`src/mim/plug/ll_nvptx/ll_nvptx.cpp:35`

```cpp
auto dot_pos = out.find('.');
assert(dot_pos < out.size());
...
auto compute_cap = std::format("{}{}", out.substr(0, dot_pos), out.substr(dot_pos + 1));
```

If `nvidia-smi` output has no `.` (or an unexpected format that still passes the all-digits loop), `dot_pos == npos`. In `NDEBUG` builds the assert vanishes; `out.substr(dot_pos + 1)` is `out.substr(0)` (npos+1 wraps to 0), so `"75"` → `compute_cap == "7575"`. That bogus value flows into `compile2ptx`/`compile2cubin` as `sm_7575`, and those failures throw plain `error(...)` (a `std::logic_error`, **not** `CmdNotFound`) that likewise escapes the fallback `catch` → abort. Validate `dot_pos != npos` explicitly and fall back to `default_compute_cap`.

---

### 5. `parse_nvcc_profile`: `find(...) == npos` not handled before `substr`
`src/mim/plug/ll_nvptx/ll_nvptx.cpp:68,72`

```cpp
size_t macro_pos = line.find("$(_HERE_)/");
top_dir          = line.substr(macro_pos + 10);   // macro_pos may be npos
...
size_t macro_pos = line.find("$(TOP)/");
lib_dir          = line.substr(macro_pos + 7);    // same
```

If a `TOP=`/`NVVMIR_LIBRARY_DIR=` line doesn't contain the expected macro, `macro_pos == npos`; `npos + 10` wraps to `9`, so `substr(9)` silently yields a wrong directory fragment rather than being detected as unparseable. **Failure scenario:** a `nvcc.profile` variant (or a relative/non-macro form) produces a nonsense libdevice path; downstream `link_libdevice` then errors on a misleading path. Guard each `find` result.

---

### 6. `split_apply`: dropping the non-`App` error silently runs the phase unconfigured
`src/mim/plug/gpu/phase/split_apply.cpp:19`

```cpp
-    auto app = body->isa<App>();
-    if (!app) error("Phase is unexpectedly not an App");
-    phase->apply(app);
+    if (auto app = body->isa<App>()) phase->apply(app);
     phase->run();
```

The removed guard used to make a non-`App` body fatal. Now such a body skips `apply()` and the phase `run()`s without ever receiving its configuration. **Failure scenario:** a stage that relies on `apply(app)` to install its arguments runs with default/empty state and produces wrong output instead of a clear error. If some phases legitimately take no args, worth a comment saying so; otherwise the previous hard failure was safer.

---

### 7. `emit_cu_error_handling` is a no-op → every CUDA driver error is ignored
`src/mim/plug/ll_nvptx/phase/ll_nvptx.cpp:136-138` (`// TODO: implement`)

Every `cuInit`/`cuCtxCreate`/`cuModuleLoadFatBinary`/`cuMemAlloc`/`cuLaunchKernel`/… result is discarded. **Failure scenario:** on device OOM, a bad fatbin, or a launch error, the generated host program continues with uninitialized handles / null pointers and silently produces wrong results (the lit tests that check `test $? -eq N` would see undefined exit codes). It's marked TODO, but flagging because it's load-bearing for correctness of the emitted programs, not just a nicety.

---

### 8. `split_off_kernels`: mutating the *old-world* lam during an `RWPhase`
`src/mim/plug/gpu/phase/split_off_kernels.cpp:42`

```cpp
old_lam->set<true>(old_lam->unique_name());
```

This writes to `old_lam` (the source world being rewritten) inside `rewrite_mut_Lam`, whereas the surrounding logic renames `new_lam`. Setting external state on the immutable source graph mid-rewrite is surprising and easy to break under re-runs/caching. **Cost:** if the intent is to give the kernel a stable external symbol, doing it on `new_lam` (which the very next lines already name via `new_lam->sym()`) would be the correct depth; mutating `old_lam` looks like it's working around name loss rather than fixing it at the source. Worth confirming this is deliberate.

---

## Notes (not counted as findings)
- `emit_epilogue` in `include/mim/plug/ll/ll.h` — the new `Axm::isa<mem::M>` guards (case 1 and the `insertvalue` loop) are effectively defensive: mem values already `emit_unsafe` to empty and are filtered before the switch, so `types` never actually contains a `mem::M`. Harmless, but if that filtering assumption ever changes, the `insertvalue` positional index `i` would be misaligned against the mem-dropped struct type from `convert(world().sigma(types))`.
- `w.lit_nat(Lit::as(type_size))` then `emit(...)` in the copy/alloc host paths re-wraps an already-literal size — minor redundancy.
- `core.mim` generalizing `%core.div`/`%core.abs` to `%mem.M a` looks intentional for addr-space work; I did not trace `normalize_div`/`normalize_abs` to confirm they don't hardcode `%mem.M 0` — worth a glance.

ftschirpke:
Thank you, I resolved the mentioned issues.
And with arguments for phases from #487, I now also made the backend customizable.
Most notably, you can now specify the compute capability via `-X ll_nvptx:sm=80` and (by default) also embed PTX in the fat binary as a JIT-compilable fallback to CUBIN.
These changes should increase portability and cross-compilation possibilities significantly.

docs/cli.md
```diff
@@ -38,6 +38,14 @@ The syntax is `-X <plugin>:<arg>`:
...
+| [ll_nvptx](@ref ll_nvptx) | `o-dev=<file>`     | Write the device LLVM IR to `<file>` instead of the default `<world>_dev.ll`/`a_dev.ll`. See @ref ll_nvptx_cli_args.                     |
+| [ll_nvptx](@ref ll_nvptx) | `no-embed`         | Don't embed compiled device binary in host LLVM IR, which is the default behavior. See @ref ll_nvptx_cli_args.       |
+| [ll_nvptx](@ref ll_nvptx) | `no-ptx-embed`     | If embedding device binary, don't embed PTX image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.            |
+| [ll_nvptx](@ref ll_nvptx) | `no-cubin-embed`   | If embedding device binary, don't embed CUBIN image into fat binary (default: PTX and CUBIN). See @ref ll_nvptx_cli_args.          |
+| [ll_nvptx](@ref ll_nvptx) | `sm=<SM>`          | If embedding device binary, compile the device binary for compute capability sm_<SM>. See @ref ll_nvptx_cli_args.                  |
+| [ll_nvptx](@ref ll_nvptx) | `libdevice=<path>` | If embedding device binary and if linking libdevice, link against the libdevice NVVM library at <path> instead of trying to find it via CUDA paths. See @ref ll_nvptx_cli_args. |
```
leissa: make sure to escape `<path>` to `` `<path>` ``. Similar to other "tags" like `<tool>` below. Just enable `MIM_BUILD_DOCS` on your end to catch these issues early.

leissa (approved):
Great work. Everything's green now 👍

PR #490, opened 2026-07-15
fix: use c++17 for hlo tests

hsqStephenZhang:


fodinabor:
Thx

fodinabor:
Why 17 tho? :)
We're using 23 for the main project

hsqStephenZhang:
it's a workable version for these two tests. it should be c++23 though, my bad.

PR #494, opened 2026-07-17
[Matrix] Fix `fold_index` recreating tuple and thus triggering expensive alpha checking.

fodinabor:
This has a similar effect on its own as checker's alpha memoization from #493. However, the alpha memoization "solves" the issue at the core level and thus is more robust versus upcoming users.


leissa (approved):
lgtm 👍

PR #499, opened 2026-07-23
plugin args. fixes #487

### LINKED ISSUE #487: Arguments for Plugins/Phases
We require a way to provide arguments to plugins or phases, especially backends.

Especially, when we want to use other tools in our backends (like I do in the NVPTX backend), a way to configure different flags for these tools would be crucial.
It would allow the user to pick optimization levels for these tools, pick specific optimizations, configure library paths, and configure targets (e.g. for cross-compilation).

leissa:
adds plug-in args via cli. this PR already ships two:

- `-X ll:o=foo.ll` to change the standard output file name of the emitted ll file
- `-X compile:aggr=on` switches the optimize pipeline to a fp loop which by default in this PR now is not looped anymore

### DIFF
--- a/include/mim/phase.h
+++ b/include/mim/phase.h
@@ -79,6 +79,10 @@ public:
     Log& log() const { return world_.log(); }
     std::string_view name() const { return name_; }
     flags_t annex() const { return annex_; }
+
+    /// Command-line arguments passed to this Phase's plugin via `-X <plugin>:<arg>`.
+    /// Derived from Phase::annex; yields an empty Vector for name-constructed Phase%s.
+    const Vector<std::string>& args();
     ///@}
 
     /// @name Fixed-Point Handling
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -15,6 +15,30 @@ Mim looks for plugins in this order:
 4. `path/to/mim.exe/../../lib/mim`
 5. `CMAKE_INSTALL_PREFIX/lib/mim`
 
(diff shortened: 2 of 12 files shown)


leissa:
@fodinabor do you still have copilot tokens or want to review yourself?

include/mim/phase.h
```diff
@@ -79,6 +79,10 @@ class Phase : public fe::RuntimeCast<Phase> {
     Log& log() const { return world_.log(); }
     std::string_view name() const { return name_; }
     flags_t annex() const { return annex_; }
+
+    /// Command-line arguments passed to this Phase's plugin via `-X <plugin>:<arg>`.
+    /// Derived from Phase::annex; yields an empty Vector for name-constructed Phase%s.
+    const Vector<std::string>& args();
```
fodinabor, line 85: Hmm.. I'm not sure if I fully understand the current name-constructed phase infra correctly.
Does the name get resolved to an "annex"-based variant at some point -> for the execution of the phase this is actually usable?
Otherwise, if this is not usable in most phases, providing this, seems like a bug just waiting to happen.

fodinabor (approved):
Apparently, I also don't have tokens left xD

Anyways, looks good to me, I just have one question regarding the name vs annex phase infra.

leissa:
> Apparently, I also don't have tokens left xD
> 
> Anyways, looks good to me, I just have one question regarding the name vs annex phase infra.

Yes, and the question is?

PR #501, opened 2026-07-25
cleanup

### LINKED ISSUE #9: Unexpected behavior for higher-order code generation
Unexpected results are produced when the generation of LLVM Code from Thorin code containing higher-order code
is attempted.

Small example:
``` 
extern "C" {
    fn g(fn(i32)->i32) -> i32;
    fn h(i32) -> i32;
}

fn main() -> i32 {
    g(h)
}
```

generates the llvm code

```
declare i8* @malloc(i64)

declare i32 @g({i32} (i32) // Error
declare i32 @h(i32)


define i32 @main() {
main_1038:
    %g_cont_1071.ret = call i32 @g({i32} (i32 @h) // Error
    br label %return_1066
    
return_1066:
    %_1077 = phi i32 [ %g_cont_1071.ret, %main_1038 ]
    ret i32 %_1077
    
}
```

Steps: `impala -emit-llvm -Othorin [file.impala]` 

The behavior occurs also at similar points where function types are used.
For instance at the allocation of a pointer slot for a function.

Observed behavior: Invalid LLVM Code is produced.
Expected behavior: The code generation should probably throw an error if unhandled higher-order code is encountered at this point.

leissa:
- cleanup
- the four big methods moved to `ll.cpp` again: 
  - `emit_bb`
  - `emit_epilogue`
  - `convert`
  -  `finalize`
  -  Bodies move verbatim (renamed to *_impl), so member access and virtual recursion are preserved. Small methods and the tiny free helpers stay inline in the header.
- The Emitter ctor resolves function pointers once via `world().driver().GET_FUN_PTR("ll", mim_ll_X)`.
- convert constants to coding conventions (`CU_LAUNCH_KERNEL` -> `Cu_Launch_Kernel` - all caps is reserved for macros)
- fix `HACK`s
- in addition to `as` and `isa` there is now `expect` which throws with an error msg on failure - even in Release
- remove silent assumptions in llvm backend with proper errors (fixes #9)
- move throwing find_cmd/run_cmd helpers to sys.h

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -240,6 +240,8 @@ Methods beginning with
 
 - `isa` behave like `dynamic_cast`: they perform a runtime check and return `nullptr` if the cast fails;
 - `as` behave more like `static_cast`: in `Debug` builds they assert, via the corresponding `isa`, that the cast is valid.
+- `expect` behave like `as`, but - instead of merely asserting in `Debug` builds and being silently unchecked in `Release` - they *always* check via the corresponding `isa` and throw a formatted exception (via mim::throwf) when the cast fails.
+  Reach for `expect` (over `as`) whenever the assumption is really a claim about the incoming IR that should surface as a proper error message rather than a `Debug`-only assertion or Release-mode undefined behavior - e.g. in backends that validate an already-lowered program.
 
 #### General Downcast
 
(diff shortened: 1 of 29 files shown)


leissa:
also tagging @ftschirpke here to take a look

ftschirpke (approved):
The changes specifically to the ll and ll_nvptx plugins look very good to me. Especially the ones in ll_nvptx solve issues and fix inconsistencies that I wanted to or should have addressed. Looks great to me!

ftschirpke:
While looking at the cleanups, another issue I wanted to address came to mind:
Currently, the nvptx lit tests are complete copies of the gpu lit test with the same name, but (obviously) other checks.
I always thought of addressing this, similar to the `lit/buffer/*_run.mim` tests, just didn't yet. Maybe this would fit great here, especially if you or @fodinabor already have experience with this, otherwise I will do this soon.

leissa:
> While looking at the cleanups, another issue I wanted to address came to mind: Currently, the nvptx lit tests are complete copies of the gpu lit test with the same name, but (obviously) other checks. I always thought of addressing this, similar to the `lit/buffer/*_run.mim` tests, just didn't yet. Maybe this would fit great here, especially if you or @fodinabor already have experience with this, otherwise I will do this soon.

we can address this in an other PR

PR #502, opened 2026-07-26
Feature/rt

### LINKED ISSUE #486: C Wrappers for LLVM backend(s)
I believe, it would be beneficial to have a pipeline that would allow us to write C wrappers to functionalities and refer to them from the backend instead of implementing them in raw LLVM IR.

The idea:
```
C wrappers --(clang)--> LLVM IR --(LL emitter)--> embedded or linked (`llvm-link`) LLVM IR
```
in order for the LL emitter to only emit a call to the wrapper function.
I think this can be especially beneficial when we emit calls to libc functions or GPU vendor functions.
For these, setup of arguments (`malloc`, `cuKernelKernel`, or future `printf`) or finding the correct implementations (CUDA driver API and `libdevice` library) can be compex to do and could be easily offloaded to the `clang` compiler.

Expected result:
- backend readability and separation of concerns (emitting LLVM vs implementing intrinsics)
- robust implementations
- less maintainability effort (offload handling of changes in API or versions to `clang`)

leissa:
add runtime in C support for llvm backend.

Fixes #486 

Depends on #501


leissa:
also tagging @ftschirpke here to take a look

src/mim/plug/ll/ll.cpp
```diff
@@ -24,19 +49,1073 @@ class Emit : public Phase {
...
         for (const auto& arg : args()) {
             world().DLOG("ll backend arg: `{}`", arg);
             if (arg.starts_with("o="))
                 path = arg.substr(2);
             else if (arg.starts_with("output="))
                 path = arg.substr(7);
+            else if (arg == "rt=embed")
+                rt = Emitter::Rt::embed;
+            else if (arg == "rt=extern")
+                rt = Emitter::Rt::ext;
```
ftschirpke, line 45: I fear this list growing further and further because the ll_nvptx backend needs to copy these (I believe).

src/mim/plug/ll/ll.cpp
```diff
@@ -24,19 +49,1073 @@ class Emit : public Phase {
...
         }
         auto ofs     = std::ofstream(path);
         auto emitter = Emitter(world(), "llvm_emitter", ofs);
+        emitter.rt_mode(rt);
+        if (rt == Emitter::Rt::embed)
+            if (auto rt_ll = read_rt_module(world())) emitter.rt_module(std::move(*rt_ll));
```
ftschirpke: How would the code look in the ll_nvptx backend, which presumably requires the rt_module for ll and its own ll_nvptx rt_module?

ftschirpke: Or maybe more specifically: This is something I just worry about needing to repeat in the ll_nvptx backend.
It works, just might cause a lot of maintainability headaches in the future.

src/mim/plug/ll/ll.cpp
leissa: Once this turns out to be a problem we can toss some software engineering on it and find a solution. But right now, it's kind of like worrying about overpopulation on mars :)

src/mim/plug/ll/ll.cpp
leissa: the current design is a first step into the direction what we want. there are obviously still some things that need to be resolved once we are juggling with several backends, runtimes, etc.

cmake/Mim.cmake
```diff
@@ -194,3 +203,63 @@ function(add_mim_plugin)
...
+    set(RT_DIR ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}/mim/rt)
+    file(MAKE_DIRECTORY ${RT_DIR})
+
+    set(RT_OUTPUTS "")
+    foreach(src ${PARSED_SOURCES})
+        get_filename_component(stem ${src} NAME_WE)
+        set(out ${RT_DIR}/${stem}.ll)
+        add_custom_command(
+            OUTPUT  ${out}
+            COMMAND ${MIM_CLANG} -S -emit-llvm -O2 ${CMAKE_CURRENT_LIST_DIR}/${src} -o ${out}
```
fodinabor: If we want to spin this further, it might become useful to consider allowing custom options here, as we might e.g. consider using this for GPU intrinsic wrappers or similar - then we'd want to be able to specify a target (triple) when building the ll already.

include/mim/plug/ll/ll.h
```diff
@@ -286,6 +311,14 @@ inline std::string Emitter::convert_ret_pi(const Pi* pi) {
 inline void Emitter::start() {
     Super::start();
 
+    // Splice the runtime wrapper module first (it carries the module's target triple/datalayout).
+    if (rt_used_ && rt_ == Rt::embed) {
+        if (rt_module_.empty())
+            fe::throwf("ll backend: `-X ll:rt=embed` needs the runtime module `mim_rt.ll`, but it "
+                       "was not found (build with clang / `MIM_BUILD_LL_RUNTIME=ON`, or use `-X ll:rt=extern`)");
+        ostream() << rt_module_ << '\n';
```
fodinabor, line 328: I read `add_mim_runtime` as allowing a number of input files that all can be used by the plugin, which is actually useful for several reasons. However, this textually copying things only works for one runtime ll file - that discrepancy seems to be lingering to confuse people.

I see two options: figure out how to copy together multiple ll files (probably error-prone).
Otherwise, I think it would make sense to do sth like `${MIM_CLANG} -S -emit-llvm -O2 ${outs} -o ${plugin}_rt.ll` to get one combined `ll_rt.ll` module.

I also see an argument for keeping the ll files apart (different specializations?) but then we'd need to be able to choose ll files..

src/mim/plug/ll/ll.cpp
```diff
@@ -773,11 +799,13 @@ std::string Emitter::emit_bb_impl(BB& bb, const Def* def) {
...
     } else if (auto q = Axm::isa<clos::alloc_jmpbuf>(def)) {
-        declare("i64 @jmpbuf_size()");
+        // The size of a `jmp_buf` is platform/libc-dependent, so it is computed by a C runtime
+        // wrapper (`rt/mim_rt.c`) rather than hard-coded here; see issue #486.
+        declare_rt("i64 @mim_jmpbuf_size()");
 
         emit_unsafe(q->arg());
         auto size = name + ".size";
-        bb.assign(size, "call i64 @jmpbuf_size()");
+        bb.assign(size, "call i64 @mim_jmpbuf_size()");
```
fodinabor, line 790: huh.. how/where was `jmpbuf_size` implemented so far?

leissa: it wasn't - there was never an in-tree implementation of jmpbuf_size XD

leissa:
@ftschirpke I've now added emit_cu_error_handling to see the system in action for ll_nvptx

include/mim/plug/ll/ll.h
leissa: good catch. I asked my AI overlord to fix that via option 2 and llvm-link

fodinabor (approved):
👍🏼

ftschirpke:
> @ftschirpke I've now added emit_cu_error_handling to see the system in action for ll_nvptx

I think this works for now as an example, but I think we should move many of the CUDA functions into the C wrappers anyways and then do this error check in a nicer way (with error message etc) there

PR #503, opened 2026-07-26
refactor plugins

leissa:
Some plugins are really old and dated. This PR uses some AI to modernize them.

I also used the opportunity to modernize lit tests and let them stick to coding conventions.


leissa:
depends on #502

PR #505, opened 2026-07-27
Refactor/implicits

leissa:
make use of implicits


src/mim/plug/regex/normalizers.cpp
```diff
@@ -2,20 +2,18 @@
...
-#include "mim/world.h"
+#include <mim/axm.h>
+#include <mim/def.h>
+#include <mim/tuple.h>
+#include <mim/world.h>
 
-#include "mim/util/dbg.h"
-#include "mim/util/log.h"
+#include <mim/util/dbg.h>
+#include <mim/util/log.h>
```
fodinabor, line 15: why the style change here?

leissa: the logic is to use `#include "..."` for includes within the own shared obj and using `#include <...>` for foreign shared objs - but we've never really enforced it very strictly through out the code base. maybe just use `#include "..."` everywhere?

fodinabor: Hmm ok.. I always had the tendency to use `".."` for "same project / repo / .." vs `<..>` anything not in the current project (like external libraries, ...)
A plugin that does not live in the main repo would also use `<mim/..>` in my usual style.

The "shared obj" variant, seems a bit harder as the question is: what lib does a header actually belong to? :)

PR #506, opened 2026-07-29
[Fix] NVPTX plugin tests use GPU plugin test code directly

ftschirpke:
The source code in the current nvptx tests is just copied 100% from gpu tests.
This PR fixes this code duplication. The nvptx plugin's tests now compile the gpu test and check/run the output.


leissa (approved):
lgtm 👍

PR #508, opened 2026-08-07
[tensor] Use mutable continutations to get fresh (distinct) memory to prevent buffer aliasing in tensor operations.

fodinabor:
Just using `bot:%mem.M 0` actually means that certain operations would be hash-consed instead of generating a fresh new allocation of a buffer. There are still a few places where allowing hash-consing might actually be okay (splat_buffer, e.g.).

### DIFF
--- a/lit/buffer/two_same_shape_binary.mim
+++ b/lit/buffer/two_same_shape_binary.mim
@@ -22,8 +22,8 @@ plugin core;
 plugin tensor;
 plugin mem;
 
-lam i32_add (a: I32, b: I32): I32 = %core.wrap.add 0 (a, b);
-lam i32_sub (a: I32, b: I32): I32 = %core.wrap.sub 0 (a, b);
+lam i32_add (a b: I32): I32 = %core.wrap.add 0 (a, b);
+lam i32_sub (a b: I32): I32 = %core.wrap.sub 0 (a, b);
 
 fun extern kernel (x y: «3, 4; I32», i j: [Idx 6, Idx 4]): I32 =
     let a  = %tensor.binary @(I32, I32, I32) i32_add @(2, (3, 4)) (x, y);
--- a/lit/buffer/two_same_shape_binary.mim
+++ b/lit/buffer/two_same_shape_binary.mim
@@ -1,0 +1,46 @@
+// RUN: %mim -p opt %s -o - > %t.out
+// RUN: %FileCheck %s < %t.out
+// RUN: %FileCheck %s --check-prefix=GONE < %t.out
+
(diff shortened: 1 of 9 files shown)


leissa:
Using a Proxy is a really bad idea. They are reserved to only use them temporarily during an analysis. Otherwise, the SEO or some other future analysis may erroneously think that your fresh mem proxy is some very specific thing in the analysis which it is not. 

A better approach would be a dedicated axiom that takes the magic number as argument. 

An even better approach would be to introduce a continuation that receives the fresh mem in the return continuation - no magic number required. Exactly the reason why [`%mem.slot`](https://mimir.github.io/mem.html#autotoc_md311) is now continuation-based:

```
ret mem = fresh_mem $ ();
```

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -216,6 +220,41 @@ const Def* LowerToMem::bot_mem() {
     return w.bot(w.call<mem::M>(0));
 }
 
+const Def* LowerToMem::fresh_mem() {
+    auto k = mem::mut_con(new_world())->set("fresh_mem");
+    pending_.push_back(k);
+    return k->var();
+}
```
leissa, line 232: can't we have a `fresh_mem` axiom in `mem.mim`? would be much cleaner:
```
ret mem = %mem.fresh $ 3; // 3 is the address space
```
in `mem.mim` would look like this:
```
axm %mem.fresh: Fn [a: Nat] → %mem.M a;
```

lit/buffer/two_same_shape_binary.mim
```diff
@@ -0,0 +1,46 @@
...
+// ga = c[i] = a[0][0] = x+y = 10+5 = 15,  gb = c[j] = b[1][2] = x-y = 30-3 = 27  =>  returns 42.
+// Aliased, both halves of `c` come from whichever loop wrote the shared buffer last: with the `add` loop
+// last, gb reads x+y = 33 instead of 27 -> 48 (the other order reads x-y = 5 for ga -> 32).
+// two_same_shape_binary_run.mim compiles this to an executable and asserts the exit code.
+plugin core;
+plugin tensor;
+plugin mem;
+
+lam i32_add (a: I32, b: I32): I32 = %core.wrap.add 0 (a, b);
+lam i32_sub (a: I32, b: I32): I32 = %core.wrap.sub 0 (a, b);
```
leissa: you can use group ptrns:
```
lam i32_sub (a b: I32): I32 = %core.wrap.sub 0 (a, b);
```

leissa (approved):

PR #510, opened 2026-08-11
fix(ll): distinguish vec.zip inputs count and lane count

hsqStephenZhang:
## Description

`select` mask for SIMD operations should match the lane count instead of inputs count.


leissa (approved):
lgtm

PR #521, opened 2026-08-20
fix(core): reduce matches on known injections

hsqStephenZhang:
## Motivation

The `torch` plugin makes extensive use of `option.unwrap_or`, and many of its option values are known at compile time. However, after inlining `unwrap_or`, matches on known union injections remain in the IR and cannot be eliminated by partial evaluation.

## Solution

Add the standard beta-reduction rule for sum types:

```text
match inject_i(v) with
  arm_0
  ...
  arm_i
  ...
= arm_i(v)
```

This is a general core reduction and allows static `option.unwrap_or` expressions to be partially evaluated without adding option-specific normalizers or lowering phases.


lit/core/match_injection.mim
```diff
@@ -0,0 +1,29 @@
+// RUN: %mim %s -o - | %FileCheck %s --implicit-check-not=Match
+
+plugin option;
```
fodinabor: should this file live in a `lit/option` folder?

hsqStephenZhang: placing under `lit/core` and `lit/option` both make sense to me. yeah, but it's more reasonable.

leissa (approved):
lgtm

PR #533, opened 2026-08-28
Preparation for GPU map_reduce

ftschirpke:
This PR implements fixes and the framework for a phase that lowers btensor map-reduce calls to GPU operations and GPU kernel launches that compute the map-reduce operation.
This PR *does not* implement this phase yet.

### DIFF
--- a/src/mim/plug/affine/phase/lower_index.cpp
+++ b/src/mim/plug/affine/phase/lower_index.cpp
@@ -123,8 +123,8 @@ const Def* LowerIndex::rewrite_imm_App(const App* app) {
     if (Axm::isa<affine::map>(app)) {
         // Extract f/idxs/sout from the *old* callee; we inline f's body at this call site rather than rewriting it into
         // a standalone lam, since its body may reference the threaded mem (from the divs) and would otherwise be open.
-        auto [mn, sinout, f, idxs] = app->callee()->as<App>()->uncurry_args<4>();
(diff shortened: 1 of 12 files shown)


src/mim/plug/affine/phase/lower_index.cpp
```diff
@@ -123,8 +123,8 @@ const Def* LowerIndex::rewrite_imm_App(const App* app) {
     if (Axm::isa<affine::map>(app)) {
         // Extract f/idxs/sout from the *old* callee; we inline f's body at this call site rather than rewriting it into
         // a standalone lam, since its body may reference the threaded mem (from the divs) and would otherwise be open.
-        auto [mn, sinout, f, idxs] = app->callee()->as<App>()->uncurry_args<4>();
-        auto [sin, sout]           = sinout->projs<2>();
+        auto [mn, sinout, f, idxs, _] = app->callee()->as<App>()->uncurry_args<5>();
+        auto [sin, sout]              = sinout->projs<2>();
 
         auto _   = Restore(mem_);
```
leissa: here is the second `_` in scope - which msvc complains about. I think this C++ feature to use `_` as throwaway var is c++26 - this is the reason why clang/gcc silently allow it.

ftschirpke: Thank you, I didn't spot it.

leissa (approved):

PR #537, opened 2026-08-29
Implement GPU lowering of map-reduce

ftschirpke:
This PR primarily implements the GPU phase LowerMapReduce. This phase lowers `%btensor.map_reduce_post` to a kernel launch with each thread executing one reduction. And the phase also emits the host operations needed to set up the data on the GPU.

Caveats:
- In general, the topic of lowering a map-reduce operation is a topic fairly new to me. Claude and I did our best to write and reuse code from the respective phase in the btensor plugin. But the code still may still require strict reviews (in the future?) @leissa
- Little testing performed
- No benchmarking performed i.e. I do not know if this currently actually yields a speedup (@HarithF I heard you did some benchmarking? Could I use this somehow?)
- We insert `%gpu.init` and `%gpu.deinit` for each map-reduce operation and try to remove as much of them as possible with a separate phase. Only inserting one each might be preferable.


src/mim/plug/gpu/phase/lower_map_reduce.cpp
```diff
@@ -1,7 +1,480 @@
...
+struct Inputs {
+    const Def* mem;
+    const Def* global;
+    DefVec dptrs;
+};
+
+Inputs alloc_copy_inputs(World& w,
+                         const Def* m0,
+                         const Def* m1,
+                         const DefVec& ris,
```
leissa: Better use `Defs` instead of `const DefVec&`.

src/mim/plug/gpu/phase/lower_map_reduce.cpp
```diff
@@ -1,7 +1,480 @@
...
+    apply_cps(w, current_mut, global_comb, {cur, elem_acc, w.tuple(input_elems)}, cont);
+
+    return kernel;
+}
+
+Lam* build_teardown(World& w,
+                    const Def* Ro,
+                    const Def* So,
+                    const Def* Tp,
+                    const DefVec& dptrs,
```
leissa: same as above

src/mim/plug/gpu/phase/lower_map_reduce.cpp
```diff
@@ -1,7 +1,480 @@
...
+    auto [post_mem, post_global, post_const] = after_launch->vars<3>();
+
+    auto [alloc_mem, host_buf] = buffer::op_alloc(Ro, So, Tp, post_mem)->projs<2>();
+    auto copy_back
+        = w.app(w.app(w.annex<gpu::buf_copy_to_host>(), {Ro, So, Tp}), {alloc_mem, post_global, out_dptr, host_buf});
+    auto [cb_mem, cb_global] = copy_back->projs<2>();
+
+    auto cur_global = cb_global;
+    for (auto dptr : dptrs)
+        cur_global = w.call(gpu::free::block, w.tuple({cur_global, dptr}));
```
leissa: more idiomatic:
```cpp
        cur_global = w.call(gpu::free::block, Defs{cur_global, dptr});
```

src/mim/plug/gpu/phase/lower_map_reduce.cpp
```diff
@@ -1,7 +1,480 @@
...
+    for (auto dptr : post_dptrs)
+        cur_global = w.call(gpu::free::block, w.tuple({cur_global, dptr}));
+    cur_global = w.call(gpu::free::block, w.tuple({cur_global, out_dptr}));
+
+    auto final_mem = w.app(w.annex<gpu::auto_deinit>(), w.tuple({cb_mem, cur_global, post_const}));
```
leissa: similar thing here:
```cpp
auto final_mem = w.app(w.annex<gpu::auto_deinit>(), {cb_mem, cur_global, post_const});
```
maybe say your clanker to look at these patterns and rewrite them.

src/mim/plug/gpu/phase/lower_map_reduce.cpp
```diff
@@ -1,7 +1,480 @@
...
+}
+
+const Def* LowerMapReduce::lower_map_reduce_post(const App* app) {
+    if (is_bootstrapping()) return Super::rewrite_imm_App(app);
+
+    auto& w = new_world();
+    auto c  = rewrite(app->callee())->as<App>();
+
+    auto [nis_nps, meta, shapes, in_tys, comb_init, acc_out, accs_all] = c->uncurry_args<7>();
+    auto [nis, nps]                                                    = nis_nps->projs<2>();
```
leissa: you can also do sth like this:
```cpp
    auto [nis, nps] = nis_nps->projs<2>([](auto d) { return Lit::isa(d); });
```
Then you directly get `nis` and `nps` as `std::optional<nat_t>`.
See similar patterns above and below.

Haven't tried it, but I think this works as well:
```cpp
auto [nis, nps] = nis_nps->projs<2>(Lit::isa);
```

leissa (approved):
just look at the nit picks, and then we can merge.

PR #551, opened 2026-09-07
feat(tensor): gather and scatter

hsqStephenZhang:
add [gather and scatter](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorgather-tensorgatherop) support for tensor plugin. 

the pseudo code is documented in `tensor.mim`

### DIFF
--- a/lit/tensor/gather_scatter.mim
+++ b/lit/tensor/gather_scatter.mim
@@ -1,0 +1,27 @@
+// RUN: %mim %s -p opt -o - | %FileCheck %s --implicit-check-not=%%tensor.gather --implicit-check-not=%%tensor.scatter
+
+// Gather lowers to indexed reads; scatter lowers to an ordered sequence of
+// reads and writes. Their element formulas remain visible as tensor get/set.
+// CHECK-DAG: fun extern gather2d
+// CHECK-DAG: fun extern scatter2d
+// CHECK: %tensor.get
+// CHECK: %tensor.set
+
+plugin core;
+plugin tensor;
+plugin affine;
+plugin compile;
+
+fun extern gather2d (input: «2, 3; I32», index: «2, 2; Idx 3»): «2, 2; I32» =
+    return (%tensor.gather @(I32, 2) ((2, 3), (2, 2)) 1_2 (input, index));
+
+fun extern scatter2d
+    (input: «2, 3; I32», index: «2, 2; Idx 3», updates: «2, 2; I32»): «2, 3; I32» =
+    return (%tensor.scatter @(I32, 2) ((2, 3), (2, 2), (2, 2)) 1_2 (input, index, updates));
+
+lam extern _compile (): %compile.Phase =
+    %compile.phases ff (
+        %tensor.lower_tensor,
+        %tensor.lower_map_reduce,
+        %compile.cleanup,
+    );
--- a/lit/tensor/gather_scatter.mim
+++ b/lit/tensor/gather_scatter.mim
@@ -13,11 +13,11 @@ plugin affine;
 plugin compile;
 
 fun extern gather2d (input: «2, 3; I32», index: «2, 2; Idx 3»): «2, 2; I32» =
-    return (%tensor.gather @(I32, 2) ((2, 3), (2, 2)) 1_2 (input, index));
+    return (%tensor.gather ((2, 3), (2, 2)) 1_2 (input, index));
 
 fun extern scatter2d
     (input: «2, 3; I32», index: «2, 2; Idx 3», updates: «2, 2; I32»): «2, 3; I32» =
-    return (%tensor.scatter @(I32, 2) ((2, 3), (2, 2), (2, 2)) 1_2 (input, index, updates));
+    return (%tensor.scatter ((2, 3), (2, 2), (2, 2)) 1_2 (input, index, updates));
 
 lam extern _compile (): %compile.Phase =
     %compile.phases ff (
--- a/src/mim/plug/tensor/phase/lower_map_reduce.cpp
+++ b/src/mim/plug/tensor/phase/lower_map_reduce.cpp
@@ -1,4 +1,5 @@
 #include "mim/plug/tensor/phase/lower_map_reduce.h"
+#include "mim/plug/tensor/phase/constraints.h"
 
 #include <optional>
 
@@ -277,7 +278,7 @@ const Def* LowerMapReduce::build_pointwise(const Def* inputs,
                                            const Def* type,
                                            const Def* So,
                                            u64 ro,
-                                           std::function<const Def*(const DefVec&, const Def*)> compute) {
+                                           std::function<const Def*(Defs, const Def*)> compute) {
     auto& w = new_world();
 
     auto fun    = w.mut_fun(inputs->type(), type)->set("pointwise");
@@ -332,7 +333,7 @@ const Def* LowerMapReduce::lower_pad(const App* app) {
     }
     auto s_out = w.tuple(so);
 
-    auto compute = [&](const DefVec& out_iters, const Def* new_inputs) -> const Def* {
+    auto compute = [&](Defs out_iters, const Def* new_inputs) -> const Def* {
         auto [input, value] = new_inputs->projs<2>();
         DefVec clamped(rn); // per-axis read index, kept in range, as `Idx (s_in#d)`
         DefVec valid;       // per-axis in-bounds flag (constant mode only)
@@ -400,7 +401,7 @@ const Def* LowerMapReduce::lower_concat(const App* app) {
         so[d] = (d == axn) ? w.lit_nat(acc_off) : Sis->proj(nisn, 0)->proj(rn, d);
     auto s_out = w.tuple(so);
 
-    auto compute = [&](const DefVec& out_iters, const Def* new_inputs) -> const Def* {
+    auto compute = [&](Defs out_iters, const Def* new_inputs) -> const Def* {
         auto o_ax = out_iters[axn];
         // Read input `i` at `out_iters`, but with the `ax` coordinate shifted by off#i and clamped into input `i`.
         auto read_i = [&](u64 i) -> const Def* {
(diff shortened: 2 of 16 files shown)


lit/tensor/gather_scatter.mim
```diff
@@ -0,0 +1,27 @@
...
+// CHECK: %tensor.get
+// CHECK: %tensor.set
+
+plugin core;
+plugin tensor;
+plugin affine;
+plugin compile;
+
+fun extern gather2d (input: «2, 3; I32», index: «2, 2; Idx 3»): «2, 2; I32» =
+    return (%tensor.gather @(I32, 2) ((2, 3), (2, 2)) 1_2 (input, index));
```
fodinabor: with the latest fixes - do we actually still have to explicitly state the `@(I32, 2)`?

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -425,6 +425,108 @@ const Def* LowerMapReduce::lower_concat(const App* app) {
...
+        current->set(true, for_call);
+        current = body;
+    }
+
+    auto step = w.app(w.annex<tensor::scatter_step_impl>(), Defs{T, r});
+    step      = w.app(step, Defs{s_src, s_idx, s_updates});
+    step      = w.app(step, dim);
+    step      = w.app(step, w.tuple(visit_indices));
+    current->app(true, cont, w.app(step, Defs{acc, index, updates}));
+    return call;
```
fodinabor: isn't this `build_elementwise`? and/or the inner part at least `build_loops`?

fodinabor (approved):
Generally looks good!

Just adding a quick review from Claude:

The element formulas and test expectations check out — gather/scatter semantics, dim-axis selection, Idx (s_src#dim) bounds, mem threading, %buffer.copy argument order, and the row-major affine.For nesting that scatter's last-write-wins depends on are all correct and consistent between the C++ lowering and the .mim _impl variants.

Four findings, roughly in order of severity:

Bounds checks are unreachable in the default pipeline — lower_map_reduce.cpp:456. The s_idx <= s_src / s_idx <= s_updates validation only lives in %tensor.lower_map_reduce, which opt.mim never runs. Via -p opt the same invalid gather bufferizes unchecked and emits an out-of-bounds %buffer.read/%mem.lea. gather_scatter_invalid.mim passes only because it drives its own pipeline.

Missing %buffer.Buf guard — lower_to_mem.cpp:699. lower_gather/lower_scatter apply %btensor.gather/scatter without checking that materialize returned a buffer. An all-size-1 tensor operand folds to a scalar never recorded in tensor_ty_, yielding an ill-typed application. lower_get/set/pad/concat/broadcast all guard this.

collect_tensor_types branches incomplete — lower_to_mem.cpp:148. The new cases omit elem(T) and the non-literal rank/axis gate(...) that pad/concat have, so a non-literal r/dim slips past the file's hard-error contract and leaves an unlowerable %btensor.gather for lower_ptr/ll.

throwf instead of nullptr on unprovable extents — lower_map_reduce.cpp:491. Every other bail-out in the phase returns nullptr; throwing here rejects shape-polymorphic gather/scatter ({n m: Nat} with distinct extent vars) that the default pipeline otherwise accepts.

Non-blocking: the two new statically_le blocks carry 8–10-line narrative comments inside braceless for bodies, against the repo's "no multi-line comments / never increase comment density" rule.

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -425,6 +425,108 @@ const Def* LowerMapReduce::lower_concat(const App* app) {
...
+        // `statically_le` can prove it (identical terms or literals); an
+        // unresolved dynamic relation is conservatively rejected here.
+        if (d != *dim_l && !statically_le(s_idx->proj(*r_l, d), s_src->proj(*r_l, d)))
+            fe::throwf("gather index shape exceeds input shape at axis {}", d);
+
+    auto compute = [&](const DefVec& out_indices, const Def* inputs) -> const Def* {
+        auto element = w.app(w.annex<tensor::gather_pointwise_elem_impl>(), Defs{T, r});
+        element      = w.app(element, Defs{s_src, s_idx});
+        element      = w.app(element, dim);
+        element      = w.app(element, w.tuple(out_indices));
```
leissa: just
```cpp
        element      = w.app(element, out_indices);
```
but you can probably just use `w.call<tensor::gather_pointwise_elem_impl>(...)` instead of stepwise manually calling a couple of times `w.app`.

There also similar spots where you can use the way more convenient `w.call` instead of hand-rolling `w.app`s.

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -425,6 +425,108 @@ const Def* LowerMapReduce::lower_concat(const App* app) {
...
+        // an input coordinate.  The index extent therefore has to fit in the
+        // corresponding source extent.  The `dim` axis is excluded because its
+        // coordinate comes from an index element of type `Idx (s_src#dim)`.
+        // This lowering currently accepts the constraint only when
+        // `statically_le` can prove it (identical terms or literals); an
+        // unresolved dynamic relation is conservatively rejected here.
+        if (d != *dim_l && !statically_le(s_idx->proj(*r_l, d), s_src->proj(*r_l, d)))
+            fe::throwf("gather index shape exceeds input shape at axis {}", d);
+
+    auto compute = [&](const DefVec& out_indices, const Def* inputs) -> const Def* {
```
leissa: `const DefVec&` -> `Defs` and similar spots in other parts of the PR

src/mim/plug/tensor/phase/lower_map_reduce.cpp
```diff
@@ -425,6 +425,108 @@ const Def* LowerMapReduce::lower_concat(const App* app) {
...
+        if (!statically_le(s_idx->proj(rn, d), s_updates->proj(rn, d)))
+            fe::throwf("scatter index shape exceeds updates shape at axis {}", d);
+    }
+
+    auto fun    = w.mut_fun(args->type(), type)->set("scatter");
+    auto ds_fun = cps::op_cps2ds_dep(fun)->set("dsFun");
+    auto call   = w.app(ds_fun, args)->set("call");
+
+    auto [input, index, updates] = fun->var(0)->projs<3>();
+    auto cont                    = fun->var(1);
```
leissa: In 99% of the cases, you *really* want to explicitly pass the arity of the extraction such as:
```cpp
    auto cont                    = fun->var(2, 1);
```
But even better:
```cpp
auto [iiu, cont] = fun->vars<2>();
auto [input, index, updates] = iiu->projs<3>();
```

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -674,4 +686,48 @@ const Def* LowerToMem::lower_concat(const App* app) {
...
+const Def* LowerToMem::lower_gather(const App* app) {
+    auto& w   = new_world();
+    auto c    = rewrite(app->callee())->as<App>();
+    auto args = rewrite(app->arg());
+
+    auto [Tr, shapes, dim] = c->uncurry_args<3>();
+    auto [T, r]            = Tr->projs<2>();
+    auto [s_src, s_idx]    = shapes->projs<2>();
+    auto [input, index]    = args->projs<2>();
+    if (!Axm::isa<buffer::Buf>(input->type()))
```
leissa: hoist `app->arg()->proj(2, 0)` and `app->arg()->proj(2, 1)` like this out of this:
```cpp
auto [x, y] = app->args<2>(); // use better names than x and y ...
```

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -674,4 +686,48 @@ const Def* LowerToMem::lower_concat(const App* app) {
...
+        input = materialize(app->arg()->proj(2, 0)->type(), app->arg()->proj(2, 0));
+    if (!Axm::isa<buffer::Buf>(index->type()))
+        index = materialize(app->arg()->proj(2, 1)->type(), app->arg()->proj(2, 1));
+
+    auto op = w.annex<btensor::gather>();
+    op      = w.app(op, Defs{T, r});
+    op      = w.app(op, Defs{s_src, s_idx});
+    op      = w.app(op, dim);
+    return w.app(op, Defs{fresh_mem(), input, index})->proj(1);
```
leissa: - again, better use `w.call` if possible. 
- `Defs{...}` -> `{...}` (at least when using `w.app`)
- `proj(1)` is dangerous. see above

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -674,4 +686,48 @@ const Def* LowerToMem::lower_concat(const App* app) {
...
+    auto [Tr, shapes, dim]           = c->uncurry_args<3>();
+    auto [T, r]                      = Tr->projs<2>();
+    auto [s_src, s_idx, s_updates]   = shapes->projs<3>();
+    auto [input, index, updates]     = args->projs<3>();
+    if (!Axm::isa<buffer::Buf>(input->type()))
+        input = materialize(app->arg()->proj(3, 0)->type(), app->arg()->proj(3, 0));
+    if (!Axm::isa<buffer::Buf>(index->type()))
+        index = materialize(app->arg()->proj(3, 1)->type(), app->arg()->proj(3, 1));
+    if (!Axm::isa<buffer::Buf>(updates->type()))
+        updates = materialize(app->arg()->proj(3, 2)->type(), app->arg()->proj(3, 2));
```
leissa: see similar comments above

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -674,4 +686,48 @@ const Def* LowerToMem::lower_concat(const App* app) {
...
+
+    auto op = w.annex<btensor::scatter>();
+    op      = w.app(op, Defs{T, r});
+    op      = w.app(op, Defs{s_src, s_idx, s_updates});
+    op      = w.app(op, dim);
+    return w.app(op, Defs{fresh_mem(), input, index, updates})->proj(1);
```
leissa: see similar comments above

PR #554, opened 2026-09-10
Add symbolic contraction support in dot product and enhance normalization logic

fodinabor:
This allows us to actually make the point that we just need to symbolically need to know that n is divisible by n, instead of requiring n to be fully known.


leissa (approved):
lgtm

PR #566, opened 2026-09-17
complete suite now uses nary extract/insert from MimIR itself

leissa:
Major points:
- Arrays and packs fuse: `«2; «3; T»»` *is* `«2, 3; T»`; a Seq carries all axes as one shape
    - Ragged nests (`«i: n, j: s#i; T»`) stay one array per axis
    - New `Shape` helper class for both Arr and Pack; Pack now has its shape as `op(0)`
- Extract and insert work on n-ary indices: `t#(i, j)` ≡ `t#i#j`
    - Size-1 axes fold away, so a 1-component `rec` sigma is the one exception
    - Rank inference falls out: `«s; T»` infers `{r, s, T}` without an explicit `@`
- `tensor.get`/`tensor.set` and the `lower_get_set` phase removed in favor of native extract/insert
- Causes a friction point similar to `ll.vec`: What exactly is a tensor?
    - Added new Annotation system:
        - These are axioms of type `{T: *} → T → T`
        - `tensor.buf` annotation that marks a term as a tensor
        - `ll.vec` and the former `clos.attr` (now `clos.anno`) are annotations
- Removed dangerous `def->proj(i)` API
- `cat_tuple`/`cat_sigma` → `Tuple::cat`/`Sigma::cat`

Here is an example:

- **old**
  ```rust
  extern fun f (input: «3, 4; I32»): «3, 4; I32» =
      let v = tensor.get ((1₃, 2₄), input);
      return (tensor.set ((0₃, 0₄), input, v));
  ```
- **new**
  ```rust
  extern fun f (input: «3, 4; I32»): «3, 4; I32» =
      let input = tensor.buf input;
      let v     = input#(1₃, 2₄);
      return (input#(0₃, 0₄) ← v); 
  ```
- **new** (means the same as above)
  ```rust
  extern fun f (input: «3, 4; I32»): «3, 4; I32» =
      let input = tensor.buf input;
      let v     = input#1₃#2₄;
      return (input#0₃#0₄ ← v);
  ```


src/mim/world.cpp
```diff
@@ -18,22 +18,78 @@ namespace mim {
...
 namespace {
 
-bool is_shape(const Def* s) {
-    if (s->isa<Nat>()) return true;
-    if (auto arr = s->isa<Arr>()) return arr->body()->zonk()->isa<Nat>();
-    if (auto sig = s->isa_imm<Sigma>())
-        return std::ranges::all_of(sig->ops(), [](const Def* op) { return op->isa<Nat>(); });
-
+/// Does `d#index \u2190 val` rebuild an aggregate right away? Mirrors the two folding rules in World::insert.
+bool insert_rebuilds(const Def* d, const Def* index, u64 threshold) {
```
fodinabor: **Insert into a fused Pack builds an ill-typed tuple.**

The scalar-unroll rule in `World::insert` (`// insert(‹4; x›, 2, y) -> (x, x, y, x)`, ~line 689) lacks the `!is_fused()` guard that this predicate and `Def::immutabilize` both carry. For a fused `‹2, 3; x›`, `arity()` is the front axis `2` and `body()` is the scalar `x`, so the rebuilt tuple mixes a row with an element. Repro, which type-checks silently:

```
extern fun f (x: I32, row: «3; I32»): «2, 3; I32» =
    let p = ‹2, 3; x›;
    return (p#0₂ ← row);
// normalizes to `return (row, x)` — second component is a scalar, not `‹3; x›`
```

Same for `‹2; ‹3; x››`, since it fuses. Suggest sharing one predicate between `insert_rebuilds` and the two inline rules; the three copies have already drifted.

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -500,41 +485,52 @@ const Def* LowerToMem::buffer_list(const Def* list, const Def* old_list, const D
...
-    auto c               = rewrite(app->callee())->as<App>();
-    auto arg             = rewrite(app->arg());
-    auto [index, arr, x] = arg->projs<3>();
-    auto [T, r, s]       = c->args<3>();
-    arr                  = to_buffer(arr, app->arg()->proj(3, 1));
-    auto buf             = Axm::isa<buffer::Buf>(arr->type());
-    if (!buf) return RWPhase::rewrite_imm_App(app); // not a recorded tensor type: leave it alone
-    auto [br, bs, bT] = buf->args<3>();             // actual (folded) buffer metadata
-    auto fidx         = fold_index(s, index);
+const Def* LowerToMem::rewrite_imm_Insert(const Insert* insert) {
```
fodinabor, line 529: **Partial tensor write is bufferized as a full-rank write.**

`rewrite_imm_Insert` has no counterpart to the `->type()->isa<Arr>()` guard on the Extract side. With `tensor.set` gone, `t#i ← row` is now expressible and reaches `buffer.write` with a rank-1 index and an array value:

```
extern fun write_row (t: «4, 8; I32», row: «8; I32»): «4, 8; I32» =
    let t = tensor.buf t;
    return (t#1₄ ← row);
```
```
error: argument is not assignable to callee's domain
  expected `[mem.M 0, buffer.Buf (2, (4, 8), I32), [Idx 4, Idx 8], I32]`,
  got      `[mem.M 0, buffer.Buf (2, (4, 8), I32), Idx 4, «8; I32»]`
```

The error also blames `affine.mim:29` rather than the user's code.

src/mim/plug/tensor/phase/lower_to_mem.cpp
```diff
@@ -500,41 +485,52 @@ const Def* LowerToMem::buffer_list(const Def* list, const Def* old_list, const D
...
+    // One Extract carries every axis, so there is no chain to walk back up and no sub-tensor to mistake the
+    // base for: an index that stops short of the element yields an array type, which the caller rejects.
+    auto ex = d->isa<Extract>();
+    if (!ex || !tensor_ty_.contains(ex->tuple()->type())) return {nullptr, nullptr};
+    return {ex->tuple(), rewrite(ex->index())};
+}
+
+const Def* LowerToMem::rewrite_imm_Extract(const Extract* extract) {
+    // A partial read yields a sub-tensor, which has no buffer of its own; `collect_tensor_types` gates on it.
+    if (is_bootstrapping() || extract->type()->isa<Arr>()) return RWPhase::rewrite_imm_Extract(extract);
```
fodinabor, line 513: **Partial tensor read produces a misleading hard error.**

`t#i` on a bufferized `«4, 8; I32»` is handed to the generic `RWPhase::rewrite_imm_Extract`, which rebuilds `Extract(buf, i)` on a `buffer.Buf`-typed operand and trips the arity check:

```
extern fun read_row (t: «4, 8; I32»): «8; I32» =
    let t = tensor.buf t;
    return (t#1₄);
```
```
affine.mim:29:20-29:23: error: index `1₄` does not fit within arity `1`
```

The comment says `collect_tensor_types` gates this case, but it contains no such gate. Either reject partial reads/writes with a proper `cannot bufferize` diagnostic or lower them to sub-buffer views; currently it is neither. A lit test for partial read, partial write, and fused-pack insert would pin this down.

src/mim/axm.cpp
```diff
@@ -11,6 +11,20 @@ Axm::Axm(NormalizeFn normalizer, u8 curry, u8 trip, const Def* type, plugin_t pl
...
 
+/// Is @p type the polymorphic identity `{T: *} → T → T`?
+static bool is_identity(const Def* type) {
+    if (auto outer = type->isa<Pi>())
+        if (auto inner = outer->codom()->isa<Pi>())
+            if (auto var = inner->dom()->isa<Var>()) return var->binder() == outer && inner->codom() == var;
+    return false;
+}
+
+const Def* Anno::isa(const Def* def) {
```
fodinabor: **Design question: `Anno::isa` is wider than the annotation set.**

Every fully-applied axm of type `{T: *} → T → T` is treated as a peelable identity annotation. That also matches `core.pe.hlt`, `core.pe.run` and `refly.check_bot`. Consequences:

- `ClosLit::fnc_as_lam` / `isa_clos_lit` (clos.cpp:42, :51) now see through a `pe.hlt` fence, so `(env, %core.pe.hlt f)` is recognized as a closure literal over `f`.
- `LowerTypedClosPrep::rewrite_imm_Tuple` still tests `Axm::isa<anno>` on the *unpeeled* operand, so the two views disagree.

If those three axioms are meant to be annotations, the `clos` matchers should be aligned. If not, `Anno::isa` needs a positive marker rather than type shape alone. Parametricity makes the type argument sound, but it does not make every identity-typed axm droppable or transparent.

src/mim/def.cpp
```diff
@@ -587,11 +588,9 @@ size_t Def::reduction_offset() const noexcept {
 
 const Def* Def::arity() const {
     switch (node()) {
-        case Node::Arr:   return op(0);
+        case Node::Arr:
+        case Node::Pack:  return Shape(op(0)).front();
```
fodinabor, line 592: **Null deref on an unset mutable Seq.** `Shape(op(0)).front()` dereferences `def_` when `op(0)` is still null (`Shape::front` → `def_->world()`). The previous version returned `nullptr` for `Arr`. `Rewriter::rewrite_mut_Seq` (rewrite.cpp:241, the `!is_set()` branch) creates exactly such a stub, so a later `num_projs()` / `var()` on it segfaults.

fodinabor (commented):
Reviewed the PR diff (base = merge base with `master`, 192 files) on today's Debug build; `ctest` (28) and `lit` (453) are green. Direction is good and the fused-shape design hangs together. Three bugs to fix before merging, one design question, a few nits — details inline.

The three bugs share one root: `tensor.get`/`set` used to reject partial access and fused packs by construction, and native `Extract`/`Insert` now let them through. Lit tests for partial read, partial write, and fused-pack insert would close the gap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

src/mim/def.cpp
```diff
@@ -625,7 +624,9 @@ const Def* Def::proj(nat_t a, nat_t i) const {
     }
 
     if (auto seq = isa<Seq>()) {
-        if (seq->has_var()) return seq->reduce(world().lit_idx(a, i));
+        auto& w = world();
+        if (seq->shape().is_fused()) return w.peel(seq, w.lit_idx(a, i)); // one axis off a fused shape yields a sub-Seq
```
fodinabor, line 628: **Not idempotent for a mutable fused Seq.** `World::peel` goes through `mut_shaped`, which mints a fresh mutable sub-Seq per call whenever the body still binds the var. Two `w.extract(x, 0₂)` calls on the same `x: «v: (2, 3); F v»` can then yield distinct `Extract` nodes with distinct types, breaking `==`-based reuse in phases.

src/mim/check.cpp
```diff
@@ -394,10 +414,17 @@ bool Checker::check1(const Seq* seq, const Def* def) {
 // Try to get rid of mut_seq's var: it may occur in its body and vanish after reduction
 // as holes might have been filled in the meantime.
 bool Checker::check(Seq* mut_seq, const Seq* imm_seq) {
-    auto mut_body = mut_seq->reduce(world().top(world().type_idx(mut_seq->arity())));
-    if (!alpha_<Check>(mut_body, imm_seq->body())) return fail<Check>();
+    // `mut_seq` binds only its own axes, so a *fused* `imm_seq` keeps the remaining ones in a sub-Seq:
+    // `«i: n; Ts#i»` against `«n, c, h; T»` matches `Ts#⊤` with `«c, h; T»`, not with `T`.
+    auto r    = mut_seq->shape().rank();
+    auto rest = r ? world().drop(imm_seq, *r) : nullptr;
```
fodinabor, line 420: This now fails outright whenever `drop` returns null, i.e. a dynamic-rank or under-ranked `imm_seq` (e.g. `imm «s; T»` with `s: «r; Nat»`, `r` still a Hole). The replaced code compared bodies directly and could still unify `Ts#⊤` with `T`. Fine if that case is unreachable, but worth a test.

src/mim/plug/btensor/phase/lower_map_reduce.cpp
```diff
@@ -281,14 +262,14 @@ const Def* LowerMapReduce::lower_map_reduce_post(const App* app) {
...
+            auto [prd_mem, p_val] = buffer::op_read(pr, ps_, pT, pcur, p_buf, *Shape(pcoords).fold(sps_j))->projs<2>();
             pcur                  = prd_mem;
             post_elems[j]         = p_val;
         }
         auto after_post            = mem::mut_con(Tp)->set("afterPost");
         auto [post_mem, elem_post] = after_post->vars<2>();
-        auto stored = buffer::op_write(obr, obs, obT, post_mem, wu, fold_index(So, write_coords), elem_post);
-        after_post->app(true, wk, w.tuple({stored->proj(0), stored->proj(1)}));
+        auto stored = buffer::op_write(obr, obs, obT, post_mem, wu, *Shape(write_coords).fold(So), elem_post);
+        after_post->app(true, wk, w.tuple({stored->proj(2, 0), stored->proj(2, 1)}));
```
fodinabor, line 272: `w.tuple({stored->proj(2, 0), stored->proj(2, 1)})` is `stored` itself (after an alpha-compare eta-reduction on every lowering), and repeated `proj(2, i)` is against the cpp-style rule (`projs<2>()` + structured binding). Same pattern in affine `lower_index.cpp:140-144`.

src/mim/plug/affine/phase/lower_index.cpp
```diff
@@ -135,7 +135,7 @@ const Def* LowerIndex::rewrite_imm_App(const App* app) {
         Lam* rw_idx_lam  = nullptr;
         if (auto idx_lam = lookup(f_lam)) {
             if (auto idx_lam_mut = idx_lam->isa_mut<Lam>();
-                idx_lam_mut && idx_lam_mut->num_vars() == 2 && idx_lam_mut->var(0)->type() == mem->type())
+                idx_lam_mut && idx_lam_mut->num_vars() == 2 && idx_lam_mut->var(2, 0)->type() == mem->type())
                 // completely rewritten, great!
                 idx_map_lam = idx_lam->as_mut<Lam>();
```
fodinabor, line 140: nit: `outs->proj(2, 1)` / `outs->proj(2, 0)` → `auto [a, b] = outs->projs<2>();` per cpp-style.

src/mim/tuple.cpp
```diff
@@ -6,13 +6,81 @@
 
 namespace mim {
 
-namespace {
-/// The App::callee of @p def - or `nullptr`, if @p def isn't an App at all.
-const Def* callee_of(const Def* def) {
-    auto app = def->isa<App>();
-    return app ? app->callee() : nullptr;
+/*
+ * Shape
```
fodinabor, line 10: nit: new `/* Shape */` banner; the repo comment rules forbid adding section-header comments (the `cat` banner below predates this PR).

src/mim/world.cpp
fodinabor: Some more context:

Failing: any row or slab insert into a fused Pack with a literal index

```
// 1. ‹2, 3; x›#0₂ ← row           →  (row, x)
// 2. ‹2, 3; x›#1₂ ← row           →  (x, row)
// 3. ‹2; ‹3; x››#1₂ ← row         →  (x, row)      nested pack fuses first, same result
// 4. ‹2, 3, 4; x›#0₂ ← slab       →  (slab, x)     slab: «3, 4; I32»
```
In every case the untouched component is the scalar body x, where a ‹3; x› row or ‹3, 4; x› slab is required. The result is stamped with the declared type «2, 3; I32» and no error is raised. The only thing that varies is the rank and the index, so the trigger is simply: fused Pack, literal index on the first axis, arity below the scalarize threshold.

Passing controls
```
// 5. ‹2, 3, 4; x›#(1₂, 2₃) ← row   →  stays a residual Insert (index is a tuple, not a Lit)
// 6. ‹2, 3; x›#(1₂, 2₃) ← y        →  stays a residual Insert
// 7. ‹3; x›#1₃ ← y                 →  (x, y, x)     non-fused, correct
// 8. (r0, r1)#0₂ ← row             →  (row, r1)     Tuple path, correct
```
Note that 5 and 6 are only safe because the n-ary index is a tuple literal and so misses the Lit::isa test. A full-rank element insert with a single-axis literal index is exactly case 7, which is fine.

src/mim/tuple.cpp
leissa: Those rules are just for the AI which spams everything with comments lol

leissa:
- `Anno`tation gone. That was more confusing then helping...
- Partial reads/writes now produce an error. In the future we might want so support subbuffer views.
  - The gap is pre-existing, not introduced here. `tensor.get`/`set` never rejected partial access "by construction" - `{T, r, s}` instantiates perfectly well at a partial rank; what caught it was elem(T) firing incidentally on the array element type. Meanwhile plain Extract/Insert on a bufferized tensor was completely unguarded on master and failed silently, which is worse than the branch's loud exception.

    So removing `get`/`set` didn't open the hole — it removed the one route that happened to trip an unrelated gate, and promoted the unguarded route to being the only way to index a tensor. That argues the gate belongs exactly where I put it, on Extract/Insert in collect_tensor_types, rather than reconstructed per-axiom: it now covers both spellings and both directions, with the user's source location.

src/mim/world.cpp
leissa: see my comment below

src/mim/plug/tensor/phase/lower_to_mem.cpp
leissa: see also below

src/mim/axm.cpp
leissa: see my comment below

src/mim/def.cpp
leissa: fixed.

src/mim/plug/btensor/phase/lower_map_reduce.cpp
leissa: fixed

src/mim/plug/affine/phase/lower_index.cpp
leissa: fixed

fodinabor:
With `Anno` gone, the annotations will not be transparent anymore for plugins not knowing the annotations of other plugins, right? I.e. there's no way left to "look through" the annotations by default?

(I'm just asking, I'm not sure that we do want to see through, as this would likely lead to the annotation being dropped relatively often. OTOH, if we don't peel it, pattern matching here and there might be inhibited.. but we'll just have to see how that works out)

leissa:
> With `Anno` gone, the annotations will not be transparent anymore for plugins not knowing the annotations of other plugins, right? I.e. there's no way left to "look through" the annotations by default?
> 
> (I'm just asking, I'm not sure that we do want to see through, as this would likely lead to the annotation being dropped relatively often. OTOH, if we don't peel it, pattern matching here and there might be inhibited.. but we'll just have to see how that works out)

Right now, no, there is no automatic peeling of annotations. And it's debatable. If those annotations simply disappear all of the place, what's the point in having them in the first place? Right now, the annotations that we have annotate variables - with which we can't do much anyway. But yes, that's a good question, where exactly do we want to place these annotations. But I guess this is a discussion for another day. We already have `ll.vec` and `clos.anno` as annotations. This PR adds `tensor.buf` to that list.

